PR #12, opened 2024-07-24
build(citizen-func): Update redis to the latest version

Garma00:
#### List of Changes

* [build: update redis to the latest version](https://github.com/pagopa/io-messages/commit/d305a7bcaced5592de63f22082b1e51eaf0d931e);
* [build: update redis import to match the new version](https://github.com/pagopa/io-messages/commit/18a7ab45b93218074e7ff0dcef2b14ed7b05a89f);
* [build: update redis utils](https://github.com/pagopa/io-messages/commit/f0e89edb6dfcdb19f066eb14ed4eb472e8217bc2);
* [test: fix unit test after redis update](https://github.com/pagopa/io-messages/commit/4fa804d0658acbd419a8ce8d372ad2b9f3ff4490).

#### Motivation and Context

We want to update redis to the latest version due to the fact that we are still using the 3 major.

Unit test I did some test locally with a redis container to ensure the new SDK behaviour, in fact i had to modify some unit test scenaios.

#### Types of changes

#### Checklist:


apps/citizen-func/utils/redis_storage.ts
```diff
@@ -2,204 +2,70 @@ import * as E from "fp-ts/lib/Either";
...
-  redisClient: RedisClient,
+  redisClient: redis.RedisClientType,
   key: string,
   value: string,
   expirationInSeconds: number,
   errorMsg?: string
 ): TE.TaskEither<Error, true> =>
   pipe(
     TE.tryCatch(
+      () => redisClient.set(key, value, { EX: expirationInSeconds }),
```
michaeldisaro: I think the new redis client have something like `setEx` command. 
Just see here https://github.com/search?q=repo%3Apagopa%2Fio-backend+setEx&type=code

michaeldisaro (changes requested):

apps/citizen-func/utils/redis_storage.ts
Garma00: @michaeldisaro If I am not mistaken the `setEx` method you are mentioning here is for redis-cluster, i passed EX as part of `SetOption` as is indicated [here](https://github.com/redis/node-redis/blob/b884e132fa54908b022f565263d1402b43cec83a/packages/client/lib/commands/SET.ts#L27)

apps/citizen-func/utils/redis_storage.ts
```diff
@@ -2,204 +2,70 @@ import * as E from "fp-ts/lib/Either";
...
-              )
-            )
-          )
-        ),
-      E.toError
+      () => redisClient.del(key),
+      () => new Error("Error deleting key value pair on redis")
     ),
-    TE.chain(TE.fromEither)
+    TE.map(_ => true)
```
lucacavallaro: here you can use `constTrue` from `fp-ts/lib/function`

```suggestion
    TE.map(constTrue)
```

apps/citizen-func/utils/redis.ts
```diff
@@ -43,7 +44,7 @@ const createClusterRedisClient = (
         port: redisPort
       }
     ]
-  }) as redis.RedisClient; // Casting RedisClustr with missing typings to RedisClient (same usage).
+  }) as redis.RedisClientType; // Casting RedisClustr with missing typings to RedisClient (same usage).
```
lucacavallaro: we should avoid this kind of casting :'( it is a bad practice, what is preventing you to return the actual redis client type?

Garma00: You are right @lucacavallaro, I also have a consideration here, this function should create a new redis cluster using an old and no longer maintained library [redis-clustr](https://www.npmjs.com/package/redis-clustr).
The correct thing to do at this point (in my opinion) would be to remove this library and use the`redis` library (we have just updated) to create a cluster BUT, i noticed that we are not (and we do not plan to) use a cluster for this function. In fact to create a cluster we should have a couple of redis nodes which we do not have as far as I now.
My proposal is to eliminate the code that deals with cluster management together with the library `redis-clustr` so we can clean up some useless code, what's your take @lucacavallaro @michaeldisaro?

apps/citizen-func/utils/redis_storage.ts
Garma00: Thanks, did it here [8b808bc](https://github.com/pagopa/io-messages/pull/12/commits/8b808bce649acdcfbf9202fcf0afdc6f93e6021c)

apps/citizen-func/utils/redis_storage.ts
michaeldisaro: You could also be right from docs, but we used `setEx` in many upgrades of redis till today https://github.com/search?q=org%3Apagopa+setEx%28&type=code 
I'd prefer to maintain coherence with other repos because some are really recent and not old upgraded once (see fast-login).

apps/citizen-func/utils/redis.ts
michaeldisaro: Be carefull, I don't think we need redis cluster anymore in this functions... we are using a single redis as a cache, we are not sharding data on a cluster.

michaeldisaro: PS: Just to be clear, I agree with removing redis cluster and using just the new redis library.

apps/citizen-func/utils/redis_storage.ts
Garma00: Ok @michaeldisaro understood, I did the fix you requested here [e5c15f3](https://github.com/pagopa/io-messages/pull/12/commits/e5c15f339a440d276dc874f2231f655e1fd79a53), also please check those changes to unit tests to ensure that the behaviour si correct [1ca5dbe](https://github.com/pagopa/io-messages/pull/12/commits/1ca5dbe30372690797adea7a968e691d79329998)

apps/citizen-func/utils/redis.ts
Garma00: Removed here [31ae191](https://github.com/pagopa/io-messages/pull/12/commits/31ae1910774b071c4d2beba290dd8ffeddc40d49)

apps/citizen-func/utils/redis_storage.ts
michaeldisaro: LGTM, new redis commands returns promises, so tests look legit to me.

Garma00: @michaeldisaro I also removed `jest-mock-express`  here [212219d](https://github.com/pagopa/io-messages/pull/12/commits/212219d27d3725011f89c0750ef5034496692466) cause it was unused as well, we got this unused dep also in `sending-func`

apps/citizen-func/utils/redis.ts
michaeldisaro: LGTM

michaeldisaro: Is the cast still necessary now?

Garma00: I removed the cast together with the `redis-clustr` and the related code.
Resolving this conversation 😄

apps/citizen-func/utils/redis.ts
```diff
@@ -1,31 +1,125 @@
...
+ *
+ * @param config
+ * @returns
+ */
+export const CreateRedisClientSingleton = (
+  config: RedisParams
+): TE.TaskEither<Error, redis.RedisClientType> =>
+  pipe(
+    TE.of(void 0),
+    TE.chainW(() =>
```
michaeldisaro: why not just TE.of(pipe(...))?

apps/citizen-func/utils/redis.ts
```diff
@@ -1,31 +1,125 @@
...
+      pipe(
+        REDIS_CLIENT,
+        TE.fromPredicate(
+          (maybeRedisCliend): maybeRedisCliend is redis.RedisClientType =>
+            maybeRedisCliend !== undefined,
+          () => void 0 // Redis Client not yet instantiated
```
michaeldisaro: constVoid

apps/citizen-func/utils/redis_storage.ts
```diff
@@ -4,67 +4,86 @@ import * as TE from "fp-ts/lib/TaskEither";
...
+      TE.tryCatch(
+        () => client.setEx(key, expirationInSeconds, value),
+        () =>
+          new Error(
+            errorMsg ? errorMsg : "Error setting key value pair on redis"
+          )
+      )
     ),
-    TE.map(_ => true)
+    TE.map(constTrue)
```
michaeldisaro: Can't we check if the response is `OK` or other?

apps/citizen-func/utils/redis_storage.ts
```diff
@@ -4,67 +4,86 @@ import * as TE from "fp-ts/lib/TaskEither";
...
+      TE.tryCatch(
+        () => client.set(key, value),
+        () =>
+          new Error(
+            errorMsg ? errorMsg : "Error setting key value pair on redis"
+          )
+      )
     ),
-    TE.map(_ => true)
+    TE.map(constTrue)
```
michaeldisaro: ditto

apps/citizen-func/utils/redis_storage.ts
```diff
@@ -4,67 +4,86 @@ import * as TE from "fp-ts/lib/TaskEither";
...
-      () => redisClient.del(key),
-      () => new Error("Error deleting key value pair on redis")
+    redisClientTask,
+    TE.chain(client =>
+      TE.tryCatch(
+        () => client.del(key),
+        () => new Error("Error deleting key value pair on redis")
+      )
     ),
     TE.map(constTrue)
```
michaeldisaro: ditto

apps/citizen-func/utils/redis_storage.ts
```diff
@@ -4,67 +4,86 @@ import * as TE from "fp-ts/lib/TaskEither";
...
-      () => redisClient.get(key),
-      () => new Error("Error while retrieving value from redis")
+    redisClientTask,
+    TE.chain(client =>
+      TE.tryCatch(
+        () => client.get(key),
+        () => new Error("Error while retrieving value from redis")
+      )
     ),
     TE.map(O.fromNullable)
```
michaeldisaro, line 64: ditto

apps/citizen-func/utils/redis_storage.ts
```diff
@@ -4,67 +4,86 @@ import * as TE from "fp-ts/lib/TaskEither";
...
-      () => new Error("Error while pinging redis")
+    redisClientTask,
+    TE.chain(client =>
+      TE.tryCatch(
+        () => client.ping("ping message"),
+        () => new Error("Error while pinging redis")
+      )
     ),
-    TE.map(_ => true)
+    TE.map(constTrue)
```
michaeldisaro, line 78: ditto

apps/citizen-func/GetMessage/handler.ts
```diff
@@ -94,7 +94,7 @@ type IGetMessageHandler = (
 const getErrorOrPaymentData = async (
   context: Context,
   serviceModel: ServiceModel,
-  redisClient: redis.RedisClientType,
+  redisClient: TE.TaskEither<Error, redis.RedisClientType>,
```
michaeldisaro: rename to `redisClientTask`?

apps/citizen-func/utils/redis_storage.ts
Garma00: In case of error we would be in the left case i think

apps/citizen-func/GetMessage/handler.ts
Garma00: [03f5440](https://github.com/pagopa/io-messages/pull/12/commits/03f5440d98abe5c1e384f39b0020591e832276c3)

apps/citizen-func/utils/redis.ts
Garma00: [fd13c0d](https://github.com/pagopa/io-messages/pull/12/commits/fd13c0d75107310eccef565c7fb3a38f42080207)

apps/citizen-func/utils/redis_storage.ts
michaeldisaro: You can see what I mean here https://github.com/pagopa/io-functions-cgn/blob/4df37c956d23a4aa2364a3017e4398bc71a81f18/utils/redis_storage.ts#L84 
Redis returns always something when in `right`.

michaeldisaro: But be careful, it does not always return `OK`, sometimes it returns a number to indicate the number of touched records, and so on... just verify. The above example has some cases.

Garma00: Ok I get what you mean, according to the [doc](https://redis.io/docs/latest/commands/setex/) it seems to me that the return value should always be `OK` for the `setEx` method so I think that we can just use `constTrue` here, I can check for others ditto.
Do you agree?

michaeldisaro: In theory, yes; but what if we get, for any reason, something different than `OK`? Are we checking it or just putting faith in it? `costTrue` is faith, an explicit check could be a good way to raise errors before finding yourself troubleshooting "why some records are not in redis and you didn't catch any error".
What do you think?

Garma00: Ok I agree, I'll add a check as you suggested

apps/citizen-func/utils/redis_storage.ts
Garma00: What kind of additional check do we expect here? I also checked the example you provided but the code seems very similar to me https://github.com/pagopa/io-functions-cgn/blob/4df37c956d23a4aa2364a3017e4398bc71a81f18/utils/redis_storage.ts#L84

apps/citizen-func/utils/redis_storage.ts
Garma00: [2270199](https://github.com/pagopa/io-messages/pull/12/commits/2270199ff45cfe82375a0f31e0540b3819f0aebb)

apps/citizen-func/utils/redis_storage.ts
Garma00: [2270199](https://github.com/pagopa/io-messages/pull/12/commits/2270199ff45cfe82375a0f31e0540b3819f0aebb)

apps/citizen-func/utils/redis_storage.ts
Garma00: [2270199](https://github.com/pagopa/io-messages/pull/12/commits/2270199ff45cfe82375a0f31e0540b3819f0aebb)

apps/citizen-func/utils/redis_storage.ts
michaeldisaro: No check, in effect. My fault.

apps/citizen-func/utils/redis_storage.ts
```diff
@@ -1,205 +1,98 @@
...
-    return E.left<Error, boolean>(err);
-  }
-  if (expectedReply !== undefined && expectedReply !== reply) {
-    return E.right<Error, boolean>(false);
-  }
-  return E.right<Error, boolean>(typeof reply === "number");
-};
-
-const isTrue = (b: boolean): b is true => b === true;
+const isSetResponseOk = (response: string): boolean => response === "OK";
```
michaeldisaro: You had `singleStringReply` in `redis.ts`.

apps/citizen-func/utils/redis_storage.ts
```diff
@@ -1,205 +1,98 @@
...
+const isDelResponseOk = (response: number): boolean => response >= 0;
```
michaeldisaro: We always used utils like these https://github.com/pagopa/io-functions-cgn/blob/4df37c956d23a4aa2364a3017e4398bc71a81f18/utils/redis_storage.ts#L25 to do this kind of check. Async or sync. I'd stick to that convention.

apps/citizen-func/utils/redis_storage.ts
michaeldisaro: We always used utils like these https://github.com/pagopa/io-functions-cgn/blob/4df37c956d23a4aa2364a3017e4398bc71a81f18/utils/redis_storage.ts#L25 to do this kind of check. Async or not.

apps/citizen-func/utils/redis_storage.ts
Garma00: [491deef](https://github.com/pagopa/io-messages/pull/12/commits/491deefe11b65f378a886272732bf3b3b00bc8de)

apps/citizen-func/utils/redis_storage.ts
Garma00: [491deef](https://github.com/pagopa/io-messages/pull/12/commits/491deefe11b65f378a886272732bf3b3b00bc8de)

apps/citizen-func/utils/redis_storage.ts
Garma00: @michaeldisaro I removed the del utility cause we never use [bac1a14](https://github.com/pagopa/io-messages/pull/12/commits/bac1a14da123aae8ad1822e0d07abeed805cb03c)

apps/citizen-func/utils/__tests__/redis_storage.test.ts
```diff
@@ -17,110 +13,73 @@ const aRedisDefaultExpiration = 10;
...
-const delMock = jest.fn().mockImplementation((_, cb) => cb(null, 1));
-const redisClientMock = {
+const setExMock = jest.fn();
+const getMock = jest.fn();
+const delMock = jest.fn();
+const redisClientMock = ({
   get: getMock,
   set: setMock,
+  setEx: setExMock,
   del: delMock
```
michaeldisaro: if we never delete you can delete mock.

michaeldisaro (approved):
lgtm

PR #18, opened 2024-09-02
docs(process-message): add sequence diagram

Garma00:
#### List of Changes

* [Add sequence diagram](https://github.com/pagopa/io-messages/commit/af14f0f7049815cd365021efe1fa72ff5daa1144);

#### Motivation and Context

We want to document this flow.

I have a few questions:
1. In addition to uploading the plant uml here, should we also upload the diagram image so it can be easily consulted? We could create a README for each function where we include relevant diagrams and describe processes.
2. Should we group these diagrams on Confluence? Should we establish a structure for our documentation?
3. I used colors based on intuition (red for errors, orange for empty responses, green for non-empty responses, and blue to indicate the end). Should we establish a legend to maintain consistency across all our diagrams?
4. In the sequence diagram, I did not use the actual names of queues, collections, and storages because I’m not sure if we want to make them public. Do you think it’s better to avoid using the real names, or should we use the correct ones?

No tests are needed.

#### Screenshots (if appropriate):

image

#### Types of changes

#### Checklist:


lucacavallaro:
@Garma00 

> In addition to uploading the plant uml here, should we also upload the diagram image so it can be easily consulted? We could create a README for each function where we include relevant diagrams and describe processes.

Yes (see https://github.com/pagopa/io-fims/ for reference)

> Should we group these diagrams on Confluence? Should we establish a structure for our documentation?

IMHO at this moment we don't have to care about Confluence documentation

> I used colors based on intuition (red for errors, orange for empty responses, green for non-empty responses, and blue to indicate the end). Should we establish a legend to maintain consistency across all our diagrams?

IMHO we should adhere to sequence diagram conventions (use `activations`, different arrow styles ecc.). I also think marking empty/not empty responses is too detailed for a sequence diagram, the focus here are the actors and their messages.

> In the sequence diagram, I did not use the actual names of queues, collections, and storages because I’m not sure if we want to make them public. Do you think it’s better to avoid using the real names, or should we use the correct ones?

The actual names are public (our infrastructure is public, published in https://github.com/pagopa/io-infra) so they can be used in these diagrams.

IMHO the best choice is use the instance name (stripped from common `prefix` and `suffix`). For example `sending-func`, `notif-func`.

cc @iwoak

Garma00:
@lucacavallaro 

> Yes (see https://github.com/pagopa/io-fims/ for reference) 

I added the `.png` here https://github.com/pagopa/io-messages/pull/18/commits/00151e16b513211249b36b9ad50e2ce340d6b405.

> IMHO we should adhere to sequence diagram conventions (use `activations`, different arrow styles ecc.). I also think marking empty/not empty responses is too detailed for a sequence diagram, the focus here are the actors and their messages.

Ok I agree, anyway I think that red for errors and green for successes helps for the diagram's readability.

> The actual names are public (our infrastructure is public, published in https://github.com/pagopa/io-infra) so they can be used in these diagrams.
> IMHO the best choice is use the instance name (stripped from common `prefix` and `suffix`). For example `sending-func`, `notif-func`.

Ok please take a look here https://github.com/pagopa/io-messages/pull/18/commits/3a50525a02af5aa98b55c142ee4b9b428c8f6e5c.

docs/process-message/process-message.puml
```diff
@@ -0,0 +1,184 @@
...
+collections profiles
+collections "message-status" as messageStatus
+collections messages
+collections "service-preferences" as servicePreferences
+collections activations
+database "message-content" as messageStorage
+
+queue "message-procesed" as ProcessedMessage
+
+CreatedMessage -> ProcessMessage : Read an item
```
lucacavallaro: Does `CreatedMessage` queue read from `ProcessMessage` participant?

I think is the opposite, `ProcessMessage` (which seems to be an Azure Function) polls `CreatedMessage` for changes

docs/process-message/process-message.puml
```diff
@@ -0,0 +1,184 @@
...
+
+alt "Invalid JSON"
+  ProcessMessage -[#red]>X CreatedMessage: Throw an error
+end
+
+ProcessMessage -> ProcessMessage : Validate the shape of the item
+
+alt "Invalid CreatedMessageEvent"
+  ProcessMessage -[#red]>X CreatedMessage: Throw an error
+end
```
lucacavallaro: In my opinion it is too detailed, maybe we can merge these steps into one (`"Validate input" -> "Throw if invalid"`)

docs/process-message/process-message.puml
```diff
@@ -0,0 +1,184 @@
...
+
+activate ProcessMessage
+ProcessMessage -> InternalStorage : Ask for a CommonMessageData
```
lucacavallaro, line 25: What is `CommonMessageData`? why does `ProcessMessage` is asking for that info? 

Tip: you can use [Notes](https://plantuml.com/sequence-diagram#:~:text=QA%2D2503%5D-,Notes%20on%20messages,-It%20is%20possible) to give context about messages

docs/process-message/process-message.puml
```diff
@@ -0,0 +1,184 @@
...
+else "Data not found"
+  InternalStorage -> ProcessMessage : None
+  ProcessMessage -[#red]>X CreatedMessage: Throw an error
+else "Data does not respect the shape CommonMessageData"
+  InternalStorage -[#red]> ProcessMessage : Decode error
+  ProcessMessage -[#red]>X CreatedMessage: Throw an error
+else "Storage returns data"
+  InternalStorage -[#green]> ProcessMessage : Data
+  deactivate InternalStorage
+end
```
lucacavallaro: When we model error handling, we don't have to list all error cases. I'd rather focus on the logic behind these checks.

For example, in this case the `core info` is: `if CoreMessageData is not found or the retrieval fails then ProcessMessage throws`

docs/process-message/process-message.puml
```diff
@@ -0,0 +1,184 @@
...
+  servicePreferences -[#red]> ProcessMessage : 5xx from cosmos
+  ProcessMessage -[#red]>X CreatedMessage : Throw an error
+else "Cosmos returns the service preferences"
+  servicePreferences -[#green]> ProcessMessage : Ok response
+  deactivate servicePreferences
+end
+
+ProcessMessage -> ProcessMessage : Calculate inbox and channels \nblocked from service preferences
+
+ProcessMessage -> ProcessMessage : Check if the sender service is SPECIAL
```
lucacavallaro, line 102: Context needed for `SPECIAL`

lucacavallaro (changes requested):
I have an additional question: what if `ProcessMessage` throws?

docs/process-message/process-message.puml
Garma00: Yess, what i mean here is that `ProcessMessage` read an item from `CreatedMessage`, what would you change here to make it clearer?

lucacavallaro: `CreatedMessage <- ProcessMessage : Read an item`

docs/process-message/process-message.puml
Garma00: Take a look here https://github.com/pagopa/io-messages/pull/18/commits/2f2fcf0c565f27212f04f9b1e547f6703ed123be

docs/process-message/process-message.puml
Garma00: Take a look here https://github.com/pagopa/io-messages/pull/18/commits/47bd1ae1ea9d402b7c009b53f0ab9867174ec96a

docs/process-message/process-message.puml
Garma00: Take a look here https://github.com/pagopa/io-messages/pull/18/commits/fff64d56f93bd609335d277cdaa40f6a6b112af6

docs/process-message/process-message.puml
Garma00: Take a look here https://github.com/pagopa/io-messages/pull/18/commits/bc7bbca35ee9e66f28d44b1e645577028e4d0e07

docs/process-message/process-message.puml
Garma00: Take a look here https://github.com/pagopa/io-messages/pull/18/commits/5c3e0462f93dcc008520dc869d8868cc83bb2728

Garma00:
> I have an additional question: what if `ProcessMessage` throws?

@lucacavallaro If the function throws an error, it will be retried up to 5 times (default configuration). If the function still fails to complete successfully after all retries, the message will be moved to the poison queue.

lucacavallaro:
> > I have an additional question: what if `ProcessMessage` throws?
> 
> @lucacavallaro If the function throws an error, it will be retried up to 5 times (default configuration). If the function still fails to complete successfully after all retries, the message will be moved to the poison queue.

@Garma00 write it in a note!

Garma00:
> @Garma00 write it in a note!

Take a look here https://github.com/pagopa/io-messages/pull/18/commits/546fba49c99564c46d389610c6d09b6c51fe4dcd

lucacavallaro (approved):

PR #26, opened 2024-09-11
feat(monitoring): add new action group

Garma00:
#### List of Changes

- Add new action group

#### Motivation and Context

We want to move some alerts in a dedicated slack channel for out domain.

terraform plan

#### Types of changes

#### Checklist:

### DIFF
--- a/infra/resources/_modules/monitoring/action_group.tf
+++ b/infra/resources/_modules/monitoring/action_group.tf
@@ -1,5 +1,5 @@
 resource "azurerm_monitor_action_group" "io_com_error_action_group" {
-  name                = "iocom-error-ag"
+  name                = "io-p-com-error-ag-01"
   resource_group_name = var.resource_group_name
   short_name          = "iocom-error"
 
--- a/infra/resources/_modules/monitoring/action_group.tf
+++ b/infra/resources/_modules/monitoring/action_group.tf
@@ -1,0 +1,11 @@
+resource "azurerm_monitor_action_group" "io_com_error_action_group" {
+  name                = "iocom-error-ag"
+  resource_group_name = var.resource_group_name
+  short_name          = "iocom-error"
+
+  email_receiver {
+    name                    = "slack"
+    email_address           = var.alert_iocom_error_notification_slack_email
+    use_common_alert_schema = true
+  }
+}
--- a/infra/resources/_modules/monitoring/action_group.tf
+++ b/infra/resources/_modules/monitoring/action_group.tf
@@ -5,7 +5,7 @@ resource "azurerm_monitor_action_group" "io_com_error" {
 
   email_receiver {
     name                    = "slack"
-    email_address           = var.alert_iocom_error_notification_slack_email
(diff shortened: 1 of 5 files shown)


Garma00:
I’ve noticed some unexpected behavior with terraform plan. I’m only adding an Action Group, and when I run the plan locally, it correctly indicates that one resource will be added. In the GitHub Action, the plan indicates that some alerts will also be added. Can you help me understand why this is happening? @lucacavallaro @christian-calabrese

lucacavallaro:
> I’ve noticed some unexpected behavior with terraform plan. I’m only adding an Action Group, and when I run the plan locally, it correctly indicates that one resource will be added. In the GitHub Action, the plan indicates that some alerts will also be added. Can you help me understand why this is happening? @lucacavallaro @christian-calabrese

I think it is because some modules points on `main` of `pagopa/dx`. The `azure_function_app` module was recently updated to include alerts by default.

We should fix module sources to a specific version (by using the commit sha) to avoid unwanted changes (cc @acuiuli)

infra/resources/_modules/monitoring/action_group.tf
```diff
@@ -0,0 +1,11 @@
+resource "azurerm_monitor_action_group" "io_com_error_action_group" {
+  name                = "iocom-error-ag"
```
lucacavallaro: We should follow [the naming convention](https://pagopa.atlassian.net/wiki/spaces/DevEx/pages/1032880258/RFC-DX-005+Naming+convention+nuove+risorse+Azure)

I omitted `location` because it default to `global`.

```suggestion
  name                = "io-p-com-error-ag-01"
```

lucacavallaro (changes requested):

infra/resources/_modules/monitoring/action_group.tf
```diff
@@ -0,0 +1,11 @@
+resource "azurerm_monitor_action_group" "io_com_error_action_group" {
```
lucacavallaro: The `_action_group` suffix is unnecessary, because this resource will always be addressed by its [address](https://developer.hashicorp.com/terraform/cli/state/resource-addressing#resource-spec) (`azurerm_monitor_action_group.io_com_error`)

```suggestion
resource "azurerm_monitor_action_group" "io_com_error" {
```

lucacavallaro (changes requested):

infra/resources/_modules/monitoring/variables.tf
```diff
@@ -0,0 +1,9 @@
+variable "resource_group_name" {
+  type        = string
+  description = "Resource group to deploy resources to"
+}
+
+variable "alert_iocom_error_notification_slack_email" {
```
lucacavallaro: In my opinion we should not tie a variable to its specific use. In the future the slack e-mail can be referenced in other places!

```suggestion
variable "io_com_slack_email" {
```

lucacavallaro (approved):

PR #28, opened 2024-09-12
Enable Application Insights via tf module input variable

lucacavallaro:
#### List of Changes

1. Removed `ai_instrumentation_key` tf variable
2. Removed `APPINSIGHTS_*` environment variables
3. Set `application_insights_connection_string` and `application_insights_sampling_percent` tf variables

#### Motivation and Context

1. `instrumentation_key` is no longer needed
2. Our tf module does not require to explicitly set `APPINSIGHTS_*` env vars to enable ai

#### Types of changes

#### Checklist:

### DIFF
--- a/infra/resources/_modules/function_app_citizen/function_app_citizen.tf
+++ b/infra/resources/_modules/function_app_citizen/function_app_citizen.tf
@@ -26,6 +26,9 @@ module "function_app_messages_citizen" {
   app_settings      = local.messages_citizen.app_settings
   slot_app_settings = local.messages_citizen.app_settings
 
+  application_insights_connection_string   = var.ai_connection_string
+  application_insights_sampling_percentage = var.ai_sampling_percentage
+
   tags = var.tags
 }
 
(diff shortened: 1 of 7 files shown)


Garma00 (commented):
I think we also should remove `APPINSIGHTS_INSTRUMENTATION_KEY` from the `config.ts` in our functions ([see here for example](https://github.com/pagopa/io-messages/blob/main/apps/citizen-func/utils/config.ts#L104)) if we don't need it anymore

lucacavallaro:
> I think we also should remove `APPINSIGHTS_INSTRUMENTATION_KEY` from the `config.ts` in our functions ([see here for example](https://github.com/pagopa/io-messages/blob/main/apps/citizen-func/utils/config.ts#L104)) if we don't need it anymore

I noticed that `APPINSIGHTS_INSTRUMENTATIONKEY` is used in our code 

https://github.com/pagopa/io-messages/blob/cd74789966913f467f3ca8e92808df94012b46fd/apps/sending-func/utils/appinsights.ts#L16-L22

so we should continue to parse it at runtime time

krusty93 (approved):

PR #31, opened 2024-09-12
Add sequence diagram for CreateMessage

lucacavallaro:
#### List of Changes

1. Add `docs/CreateMessage/create-message.plantuml`

create-message


docs/CreateMessage/create-message.plantuml
```diff
@@ -0,0 +1,91 @@
...
+
+break Invalid //FiscalCode//
+  fn --x sender: Bad request
+end
+
+opt //Remoted Content//
+  break Invalid //configuration_id//
+    fn --x sender: Forbidden
+  end
+end
```
Garma00: Here we should have an API call to the `getRCConfiguration`, do we want to avoid it to simplify the diagram? The status we return to the caller varies based on the response from this call.

docs/CreateMessage/create-message.plantuml
```diff
@@ -0,0 +1,91 @@
...
+actor sender as "Sender"
+control fn as "CreateMessage"
+database db as "messages"
+collections st as "processing-message"
+participant ai as "ai-common"
+queue q as "message-created"
+
+sender -> fn ++: //Submit Message//
+
+fn -> fn: Parse the //Request Payload//
```
Garma00, line 16: Before this parsing, we should have a series of middleware that check if the sender has the right permissions to send a message with this payload (for example, if they can send advanced messages).

lucacavallaro: @Garma00 where are these middlewares defined? I found only one check on ACLs (https://github.com/pagopa/io-functions-services/blob/master/CreateMessage/handler.ts#L471) after the `create messageId` step

Garma00: [Here](https://github.com/pagopa/io-functions-services/blob/master/CreateMessage/handler.ts#L606) they are

lucacavallaro: thanks!!

do you know why these acls are checked two times?

Garma00: Those are 2 different checks: the `canWriteMessage` ensures that the sender is allowed to send a message to this specific citizen, while the middlewares I linked, check that the sender can send a message with that specific payload

docs/CreateMessage/create-message.png
Garma00, line 1: @lucacavallaro I think you forgot to regenerate this `png` after the changes

lucacavallaro: 71889c5bd52685490d247e19b6721aab4c1235be

Garma00 (approved):

PR #39, opened 2024-09-24
feat(etl-func): init setup etl function app

acuiuli:
#### List of Changes

- [init setup etl-func and health check](https://github.com/pagopa/io-messages/pull/39/commits/df067f44151ee34000e875617eb050eb0e30aa36)
- [added tf config for etl function app](https://github.com/pagopa/io-messages/pull/39/commits/9b5283fad6b478e4e6da49b1ea94669eb82e425b)
- [added deploy pipeline for etl-func](https://github.com/pagopa/io-messages/pull/39/commits/20c7dc5c5ec3534bf14902d578d5d88d1d877c8f)

#### Motivation and Context

Setup of a new function app called etl-func that will contains all the functions and logic for the data plan activity

#### Types of changes

- [x] Chore (nothing changes by a user perspective)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #40, opened 2024-09-25
chore: add message-status and message avro schemas

Garma00:
This PR is based on https://github.com/pagopa/io-messages/pull/39.

#### List of Changes

* Added message-status and message avro schema;
* Added tlf-func inside `labeler.yml`.

#### Motivation and Context

We need a brand new pair of `avro` schemas to describe the `message` and `message-status` shape.
Here you can find the older schemas:

- [message-status](https://github.com/pagopa/io-functions-elt/blob/main/avro/message-status.json);
- [message](https://github.com/pagopa/io-functions-elt/blob/main/avro/message.json).

Also here you can find a detailed list of changes with the older schemas:

##### message-status

- Added `schema_version`;
- Removed `DELETE` from `op`;
- Removed `default` from `id`;
- Renamed `messageId` in `message_id`;
- Removed `default` from `messageId`;
- Removed `createdAt` (redundant with `timestamp`);
- Renamed `version` in `status_version`;
- Removed `default` from `status_version`;
- Removed `THROTTLED` from `status`;
- Removed `default` from `isRead`;
- Renamed `isRead` from `is_read`;
- Removed `default` from `isArchived`;
- Renamed `isArchived` from `is_archived`

##### message

- Added `schema_version`;
- Removed `DELETE` and `UPDATE` from `UPDATE`;
- Removed `default` from `id`;
- Removed `default` from `senderServiceId`;
- Renamed `senderServiceId` in `sender_service_id`;
- Removed `default` from `senderUserId`;
- Renamed `senderUserId` in `sender_user_id`;
- Renamed `isPending` in `is_pending`;
- Removed `timeToLiveInSeconds` (no longer used);
- Removed `createdAt` (redundant with `timestamp`);
- Removed `default` from `content_subject`;
- Renamed `content_paymentData_amount` in `content_payment_data_amount`;
- Renamed `content_payment_data_notice_number` in `content_payment_data_notice_number`;
- Renamed `content_paymentData_invalidAfterDueDate` in `content_payment_data_invalid_after_due_date`;
- Renamed `content_paymentData_payeeFiscalCode` in `content_payment_data_payee_fiscal_code`;
- Renamed `featureLevelType` in `feature_level_type`;

The [notification-status](https://github.com/pagopa/io-functions-elt/blob/main/avro/notification-status.json) is defined but is not used so we can omit it here.

#### Types of changes

#### Checklist:


apps/etl-func/avro/message.json
```diff
@@ -0,0 +1,88 @@
...
+      "name": "content_type",
+      "type": [
+        "null",
+        {
+          "type": "enum",
+          "name": "MessageContentType",
+          "symbols": ["GENERIC", "PAYMENT", "EU_COVID_CERT", "LEGAL", "PN"]
+        }
+      ],
+      "default": null
```
lucacavallaro: Can you explain the meaning of these `content_type` values?
What's the difference between `null` and `GENERIC`?
As per requirements, we have to include also `PAGOPA_RECEIPT`

apps/etl-func/avro/message.json
```diff
@@ -0,0 +1,88 @@
...
+      "name": "sender_user_id",
+      "type": "string"
+    },
+    {
+      "name": "is_pending",
+      "type": "boolean",
+      "default": true
+    },
+    {
+      "name": "content_subject",
```
lucacavallaro: why are we prefixing some fields with `content_` ?

apps/etl-func/avro/message.json
```diff
@@ -0,0 +1,88 @@
...
+    },
+    {
+      "name": "content_payment_data_amount",
+      "type": "int",
+      "default": 0
+    },
+    {
+      "name": "content_payment_data_notice_number",
+      "type": "string",
+      "default": "undefined"
```
lucacavallaro: Why `"undefined"`?

apps/etl-func/avro/message.json
```diff
@@ -0,0 +1,88 @@
...
+      ],
+      "default": null
+    },
+    {
+      "name": "content_payment_data_amount",
+      "type": "int",
+      "default": 0
+    },
+    {
+      "name": "content_payment_data_notice_number",
```
lucacavallaro: what is the meaning of this field?

lucacavallaro (changes requested):

apps/etl-func/avro/message.json
Garma00: @lucacavallaro in the[ avro specification](https://avro.apache.org/docs/1.12.0/specification/) we got `doc` to put a simple description of the field, I was thinking about adding it to all the fields, what's your take?

lucacavallaro: 👍

apps/etl-func/avro/message.json
Garma00: As of today, both the subject and the message content are logically treated as part of the overall 'content' of the message, which is why, in the older schema, it was labeled 'content_subject.' I don't have a strong opinion on this, so if you think we can remove the prefix, I'm fine with that.

apps/etl-func/avro/message.json
Garma00: At this level we can't discriminate if the message is a payment message so this field should be optional

lucacavallaro: so the default value should be `null` (?)

```suggestion
      "name": "content_payment_data_notice_number",
      "type": ["null", "string"],
      "default": null
```

apps/etl-func/avro/message.json
Garma00: Take a look here https://github.com/pagopa/io-messages/pull/40/commits/efe0b201c2734690a92eede44b2e3d47e28d7f07, https://github.com/pagopa/io-messages/pull/40/commits/2ec816b6fa80b4203aab229e2e160cda65e473de

apps/etl-func/avro/message.json
Garma00: https://github.com/pagopa/io-messages/pull/40/commits/cd2ada3bc076f086566273d584282a97d491c7ec

apps/etl-func/avro/message.json
Garma00: I have added a description here https://github.com/pagopa/io-messages/pull/40/commits/efe0b201c2734690a92eede44b2e3d47e28d7f07

apps/etl-func/avro/message.json
```diff
@@ -0,0 +1,101 @@
...
+    },
+    {
+      "name": "content_type",
+      "doc": "Indicates the type of the message.",
+      "type": [
+        "null",
+        {
+          "type": "enum",
+          "name": "MessageContentType",
+          "symbols": ["GENERIC", "PAYMENT", "EU_COVID_CERT", "LEGAL", "PN"]
```
acuiuli: - Can we remove `EU_COVID_CERT` and `LEGAL`? Are they still used? 
- `PN` and `PAGOPA_RECEIPT` are not really message type, why don't we split this field into 2 and move these values into another field?

lucacavallaro: Maybe we can change the name of this field - its purpose is to "group" messages based on its content
It is not an actual "message type"

We have to review all the categories to make sure that they are correct and actually used

Garma00: I don't think that we can remove any of the symbols here even if they are no longer used cause we are going to process all the messages.
I agree that "message type" is not a proper term here, we can rename this field to something like "message-category" @lucacavallaro @acuiuli

acuiuli: You are right about removing values.

I still think that splitting this into 2 fields is the better choice. What happens when `PN` sends a payment message? What value do you choose?

lucacavallaro: @acuiuli In that case, the value should remain `PN`. The purpose of this field is to map a message to a "business/initiative category". We want to track how many messages are sent by PN, how many are pagoPA receipts, etc.

acuiuli: I thought that we also want to track how many are generic messages and how many contains a payment, with only one field we will never be able to do that accurately because a payment message sent by PN will not be counted as a payment message.
I think that we need to know if what I'm saying is a requirement or we can just ignore it and let this field as it is and just modify it's name.

lucacavallaro: > I thought that we also want to track how many are generic messages and how many contains a payment

To count how many contains a payment we can check the other (payment related) fields or maybe we can add another field called `has_payment`. 

Our only requirement right now is to include a "business category" to distinguish between different initiatives.

Garma00: So do we agree on renaming this to `message-category`? Do you think it's a good compromise? @acuiuli @lucacavallaro

lucacavallaro: I think we should involve the rest of the team tomorrow morning to decide both the name (`category` is ok?) of the field and its content.

apps/etl-func/avro/message.json
```diff
@@ -0,0 +1,138 @@
...
+      "doc": "A boolean that indicates whether the message contains sensitive informations. If this value is true the push notification sent is anonymous and the email notification will not be sent.",
+      "type": "boolean",
+      "default": false
+    },
+    {
+      "name": "timestamp",
+      "doc": "A timestamp that indicates when the message was created.",
+      "type": "long",
+      "logicalType": "timestamp-millis",
+      "default": 0
```
Garma00: Do we agree on that? or we just want to ensure that this field is valued correctly? @lucacavallaro @acuiuli

lucacavallaro: no, I think it should be always valid

apps/etl-func/avro/message.json
```diff
@@ -0,0 +1,141 @@
...
+          ]
+        }
+      ],
+      "default": null
+    },
+    {
+      "name": "content_payment_data_amount",
+      "doc": "An integer that indicates the amount of the payment.",
+      "type": "int",
+      "default": 0
```
Garma00: I was considering replacing the default value with `null`, as we did for `content_payment_data_notice_number`. If a message doesn't contain payment information, I believe it's better to avoid including any payment-related data.
If you agree with me I would do the same for `content_payment_data_invalid_after_due_date`.
@lucacavallaro @acuiuli

apps/etl-func/avro/message.json
Garma00: I agree, I'm removing the default value.

apps/etl-func/avro/message.json
lucacavallaro:
```suggestion
      "name": "content_payment_data_amount",
      "doc": "An integer that indicates the amount of the payment.",
      "type": ["null", "int"],
      "default": null
```

apps/etl-func/avro/message.json
```diff
@@ -0,0 +1,141 @@
...
+      "name": "content_payment_data_notice_number",
+      "doc": "An id that identifies the the payment.",
+      "type": ["null", "string"],
+      "default": null
+    },
+    {
+      "name": "content_payment_data_invalid_after_due_date",
+      "doc": "A boolean that indicates if the payment is considered invalid after the due date.",
+      "type": "boolean",
+      "default": false
```
lucacavallaro:
```suggestion
      "name": "content_payment_data_invalid_after_due_date",
      "doc": "A boolean that indicates if the payment is considered invalid after the due date.",
      "type": ["null", "boolean"]
      "default": null
```

apps/etl-func/avro/message.json
```diff
@@ -0,0 +1,141 @@
...
+      "doc": "A field that indicates whether the message is PREMIUM or not.",
+      "type": {
+        "type": "enum",
+        "name": "MessageFeatureLevelType",
+        "symbols": ["STANDARD", "ADVANCED"]
+      },
+      "default": "STANDARD"
+    },
+    {
+      "name": "anonymous_fiscal_code",
```
lucacavallaro: I don't like the name of this field. It is explicit, and right now it is also correct - but we are sure that in the future we will always refers `receivers` by fiscal code?

Maybe in the future we will use opaque ids o ANPR id. 

IMHO we should pick a "generic" name.

apps/etl-func/avro/message.json
```diff
@@ -0,0 +1,141 @@
...
+      "doc": "Tokenized fiscal code of the citizen.",
+      "type": "string"
+    },
+    {
+      "name": "configuration_id",
+      "doc": "An ULID that identifies the remote content configuration.",
+      "type": "string"
+    },
```
lucacavallaro: why is this value needed? I don't think that DL will use RC configurations

apps/etl-func/avro/message.json
acuiuli: I agree that there should not be any payment field if it's not a payment message

apps/etl-func/avro/message.json
Garma00: That's a good point, what about `anonymous_receiver_id`?

apps/etl-func/avro/message.json
Garma00: I removed the prefix [here](https://github.com/pagopa/io-messages/pull/40/commits/1ac5f049c9bec589693559028d0e4307d4135d4e)

apps/etl-func/avro/message-status.json
```diff
@@ -0,0 +1,63 @@
...
+    {
+      "name": "message_id",
+      "doc": "An ULID that identifies the message inside the message collection.",
+      "type": "string"
+    },
+    {
+      "name": "status_version",
+      "doc": "An integer that indicates the version of the status.",
+      "type": "int"
+    },
```
lucacavallaro, line 35: what is the purpose of this field? it points to the latest status available?

apps/etl-func/avro/message.json
lucacavallaro: I think that `receiver_id` (or `recipient_id` I don't know which is better) it's ok

lucacavallaro (changes requested):

apps/etl-func/avro/message-status.json
Garma00: message-status is a versioned collectionn, so this value indicates the current version of the document

apps/etl-func/avro/message.json
Garma00: Removed [here](https://github.com/pagopa/io-messages/pull/40/commits/bac56ac4a53445ae6bcaee77c71a10ff5590b188)

apps/etl-func/avro/message-status.json
lucacavallaro: and why it is needed? should we push a new updated `message` to the queue for every status update? 🤯

Garma00: Aren't we planning to use cosmos change feed on message-status collection? The function would be triggered for every status update

lucacavallaro: Ok, but I was expecting that for every change in `message-status`, we would enqueue a new `message-status` event and not a `message` one.

Garma00: in fact this is the message-status avro

lucacavallaro (approved):

PR #42, opened 2024-09-27
docs(fn-etl): add extract-message sequence

Garma00:
This PR is based on #39 

#### List of Changes

* Add a new directory `docs/` inside the `etl-func`;
* Add a new sequence diagram describing the wanted behavior  of this function;

#### Motivation and Context

I added this new sequence so we can use it in the DR and we can discuss about the expected behavior.

#### Types of changes

#### Checklist:


acuiuli:
Why didn't you put the diagrams inside the docs folder in the root?

Garma00:
> Why didn't you put the diagrams inside the docs folder in the root?

I believe that is better to separate the documentation of the apps but we can discuss about this

acuiuli:
> > Why didn't you put the diagrams inside the docs folder in the root?
> 
> I believe that is better to separate the documentation of the apps but we can discuss about this

I prefer the `docs` folder in the root but we need to decide on a standard and stick to it imho. Since we have also documentations that, right now, is not associated to an app (since io-fn-services is not migrated) I fear that we will end up with documentations scattered all over the place

Garma00:
I will open another PR based on the main branch soon where I'll add also the business logic.

PR #48, opened 2024-10-04
making loadConfigFromEnvironment a common package

lucamancosu:
#### List of Changes

Created a new workspace io-messages-common. The loadConfigFromEnvironment function is now a common package. 
#### Motivation and Context

Avoid code dupplication

1. root of the project, execute: yarn install
2.  root of the project, execute: yarn build
3. go inside the apps/notif-func and create a local.settings.json
4. if you put the mandatory env var inside the local.settings.json and run the yarn start command the function starts correctly
5. if you DONT put the mandatory env var inside the local.settings.json and run the yarn start command the function returns an error and a logging of the missing variables
6. Unit tests have been written

#### Types of changes

#### Checklist:


packages/io-messages-common/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "@types/node": "^20.12.4",
+    "@vitest/coverage-v8": "^1.5.0",
+    "eslint": "8.57.0",
+    "prettier": "^3.2.5",
+    "tsc-alias": "^1.8.10",
+    "typescript": "5.4.3",
+    "vitest": "^1.5.0"
+  },
+  "dependencies": {
+    "io-ts": "^2.2.21",
```
Garma00: Do we need `io-ts` as dependency here?

packages/io-messages-common/src/adapters/tests/index.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    const functionEntryPoint = vi.fn(async () => Promise.resolve());
+    const envZodSchema = z.object({
+      APPLICATIONINSIGHTS_CONNECTION_STRING: z.string().min(1),
+    });
+    vi.stubEnv("APPLICATIONINSIGHTS_CONNECTION_STRING", "stringTest");
+
+    await expect(
+      loadConfigFromEnvironment(functionEntryPoint, envZodSchema),
+    ).resolves.toEqual(undefined);
+    await expect(functionEntryPoint).toHaveBeenCalledOnce();
```
Garma00: We do not need to `await` here, also if you remove those 2` await` yo can remove the `async` in the it callback as well.
If you agree you can make the same changes to the other tests 😄

Garma00 (changes requested):
Well done, just some small changes 💯

packages/io-messages-common/src/adapters/tests/index.spec.ts
lucamancosu: loadConfigFromEnvironment is an async function. I think we must await to test how it resolves. We are awaiting for its execution in the functions codebase too. Can you explain why i must remove the await and if you can propose a way to test it without awaiting ?

packages/io-messages-common/package.json
lucamancosu: no needed

packages/io-messages-common/src/adapters/tests/index.spec.ts
Garma00: You are right but here you are using `.resolves` which is a property of `Assertion<Promise<void>>`
immagine

lucacavallaro: @Garma00, @lucamancosu is right, we need `await` here even if we use `resolves`

> If the assertion is not awaited, then you will have a false-positive test that will  pass every time. To make sure that assertions are actually called, you may use [expect.assertions(number)](https://vitest.dev/api/expect.html#expect-assertions).

https://vitest.dev/api/expect.html#resolves

Garma00: Ok sorry for the confusion, my bad!

Garma00: Sorry for reopening this thread, I just realised that you are both right on the `await` that we are using on `resolves`, on the other hand the `await` used on `toHaveBeenCalled` is not needed (also ts-server states this)
image
@lucacavallaro @lucamancosu

lucamancosu: @Garma00 you're right. Changes already pushed

Garma00 (approved):

PR #49, opened 2024-10-08
feat(fn-etl): tokenize-fiscal-code

acuiuli:
#### List of Changes

- New sequence diagram describing the behavior of the TokenizeFiscalCode function;
- TokenizeFiscalCode business logic implementation

#### Types of changes

#### Checklist:


apps/etl-func/docs/tokenize.puml
```diff
@@ -0,0 +1,29 @@
+@startuml TokenizeFiscalCode
+
+participant caller
```
lucacavallaro: imho we can remove this actor

apps/etl-func/docs/tokenize.puml
```diff
@@ -0,0 +1,29 @@
...
+participant TokenizeFiscalCode as tf
+participant PDV
+participant "ai-common" as ai
```
lucacavallaro: are we sure that we want to log errors on AI?

apps/etl-func/docs/tokenize.puml
```diff
@@ -0,0 +1,29 @@
...
+
+caller -> tf : call
+activate tf
+
+tf -> tf : Validate fiscal code
```
lucacavallaro: this procedure can accept directly a `FiscalCode` parameter, already validated

apps/etl-func/docs/tokenize.puml
```diff
@@ -0,0 +1,29 @@
...
+  tf -> caller : Throw an error
+end
+
+tf -> PDV : Call tokenizer api
+
+alt Error from PDV
+  PDV -[#red]> tf : Error
+  tf -> ai : Track the error event
+  tf -> caller : Throw an error
+end
```
lucacavallaro: no retries here?

lucacavallaro (changes requested):

apps/etl-func/docs/tokenize.puml
acuiuli: I think that having a caller helps. I modified it as an actor instead of a participant

apps/etl-func/docs/tokenize.puml
acuiuli: maybe it's better to just return an error and leave the log to the caller

apps/etl-func/src/adapters/pdv-tokenizer/pdv-tokenizer-client.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const delay = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
+
+export default class PDVTokenizerClient implements TokenizerClient {
+  #apiKey: string;
+  #basePath = "tokenizer/v1";
+  #baseUrl = "https://api.tokenizer.pdv.pagopa.it/";
+
+  constructor(apiKey: string) {
+    if (!apiKey) throw new Error("Api key is required");
```
Garma00: Is this check needed? We are accepting `string` in input and not `string | undefined`, also the Api key env var is required

apps/etl-func/src/adapters/pdv-tokenizer/pii-resource.ts
```diff
@@ -0,0 +1,4 @@
+import { z } from "zod";
+
+export const piiResourceSchema = z.object({ pii: z.string().min(1) });
```
Garma00: Why do we need this to be an object? Can't we just use a string?
```suggestion
export const piiResourceSchema = z.string().min(1);
```

apps/etl-func/src/adapters/pdv-tokenizer/token-resource.ts
```diff
@@ -0,0 +1,6 @@
+import { z } from "zod";
+
+export const tokenResourceSchema = z.object({
+  token: z.string().uuid().min(1),
+});
```
Garma00: Ditto, this could be just a string, also I think  we do not need to specify `min(1)` if this is an UUID
```suggestion
export const tokenResourceSchema = z.string().uuid();
```

apps/etl-func/src/domain/interfaces/tokenizer.ts
```diff
@@ -0,0 +1,18 @@
+import { z } from "zod";
+
+export interface TokenizerClient {
+  tokenize(pii: string): Promise<string>;
+}
+
+const messageTokenizedSchema = z.object({
+  fiscalCode: z.string().uuid().min(1),
```
Garma00: is `uuid()` correct here?

apps/etl-func/src/domain/interfaces/tokenizer.ts
```diff
@@ -0,0 +1,18 @@
...
+});
+
+export type messageTokenized = z.infer<typeof messageTokenizedSchema>;
```
Garma00:
```suggestion
export type MessageTokenized = z.infer<typeof messageTokenizedSchema>;
```

apps/etl-func/src/domain/interfaces/tokenizer.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const tokenize =
+  (message: { fiscalCode: string }) =>
+  async (client: TokenizerClient): Promise<messageTokenized> => {
+    const tokenizedFiscalCode = await client.tokenize(message.fiscalCode);
+    return { ...message, fiscalCode: tokenizedFiscalCode };
```
Garma00: Is it correct that we have logic inside `interfaces/` ?

Garma00 (changes requested):

apps/etl-func/src/adapters/pdv-tokenizer/pdv-tokenizer-client.ts
acuiuli: Even if the parameter is a string you can pass an empty string, this check avoids that

apps/etl-func/src/adapters/pdv-tokenizer/pii-resource.ts
acuiuli: this is the request of the pdv tokenizer api and it must be an object with a field named pii that is a string. You can check the openapi [here](https://api.tokenizer.pdv.pagopa.it/docs/tokenizerpapis/openapi.json)

apps/etl-func/src/adapters/pdv-tokenizer/token-resource.ts
acuiuli: Ditto. This is the response of the pdv tokenizer api and it is an object that contains a field name token that is a uuid. I agree that we can remove the min(1) thought

apps/etl-func/src/domain/interfaces/tokenizer.ts
acuiuli: The token that the tokenizer api returns is a uuid so the tokenized fiscalCode in the message will be a uuid

apps/etl-func/src/adapters/pdv-tokenizer/pdv-tokenizer-client.ts
```diff
@@ -0,0 +1,57 @@
...
+    this.#apiKey = apiKey;
+  }
+
+  async tokenize(pii: string): Promise<string> {
+    for (let attempt = 0; attempt < 3; attempt++) {
```
lucacavallaro: I think it's better to separate the `tokenize` function from the `retry logic`

apps/etl-func/src/adapters/config.ts
```diff
@@ -7,12 +7,22 @@ const logger = pino({
   level: "error",
 });
 
-export const configSchema = z.object({});
+export const configSchema = z.object({
+  pdv: z.object({
+    tokenizerApiKey: z.string().min(1),
+  }),
```
lucacavallaro:
```suggestion
  pdvTokenizer: z.object({
    apiKey: z.string().min(1),
  }),
```

apps/etl-func/src/adapters/pdv-tokenizer/pdv-tokenizer-client.ts
```diff
@@ -0,0 +1,57 @@
...
+import { piiResourceSchema } from "./pii-resource.js";
+import { problemSchema } from "./problem.js";
+import { tokenResourceSchema } from "./token-resource.js";
+
+const delay = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
+
+export default class PDVTokenizerClient implements TokenizerClient {
+  #apiKey: string;
+  #basePath = "tokenizer/v1";
+  #baseUrl = "https://api.tokenizer.pdv.pagopa.it/";
```
lucacavallaro: this url should not be hardcoded

apps/etl-func/src/adapters/pdv-tokenizer/pdv-tokenizer-client.ts
```diff
@@ -0,0 +1,57 @@
...
+
+import { piiResourceSchema } from "./pii-resource.js";
+import { problemSchema } from "./problem.js";
+import { tokenResourceSchema } from "./token-resource.js";
+
+const delay = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
+
+export default class PDVTokenizerClient implements TokenizerClient {
+  #apiKey: string;
+  #basePath = "tokenizer/v1";
```
lucacavallaro: no need to have `basePath` as property, since is not used in other methods (and it be part of the base url)

apps/etl-func/src/adapters/pdv-tokenizer/pdv-tokenizer-client.ts
lucacavallaro: prefer use `node:assert` for this kind of checks

apps/etl-func/src/adapters/pdv-tokenizer/problem.ts
```diff
@@ -0,0 +1,21 @@
...
+
+// Define the Problem schema
+export const problemSchema = z.object({
+  detail: z.string().optional(),
+  instance: z.string().optional(),
+  invalidParams: z.array(invalidParamSchema).optional(),
+  status: z.number().int(),
+  title: z.string(),
+  type: z.string().optional(),
+});
```
lucacavallaro: are we using all these fields?

apps/etl-func/src/adapters/pdv-tokenizer/token-resource.ts
```diff
@@ -0,0 +1,6 @@
+import { z } from "zod";
+
+export const tokenResourceSchema = z.object({
+  token: z.string().uuid().min(1),
```
lucacavallaro:
```suggestion
  token: z.string().uuid(),
```

apps/etl-func/src/domain/interfaces/tokenizer.ts
```diff
@@ -0,0 +1,18 @@
+import { z } from "zod";
+
+export interface TokenizerClient {
+  tokenize(pii: string): Promise<string>;
+}
```
lucacavallaro, line 3: maybe we can use more generic terms here

what about `maskSensitiveInfo`? in the future we can use these function also for masking sensitive info from `subject`

apps/etl-func/src/domain/interfaces/tokenizer.ts
```diff
@@ -0,0 +1,18 @@
...
+
+const messageTokenizedSchema = z.object({
+  fiscalCode: z.string().uuid().min(1),
+});
+
+export type MessageTokenized = z.infer<typeof messageTokenizedSchema>;
```
lucacavallaro: we tokenize the fiscalCode that is a property of the Message, not the Message itself

I would call it `MessageWithoutPII` o `MessageWithoutSensitiveData`

lucacavallaro (changes requested):

apps/etl-func/src/adapters/pdv-tokenizer/problem.ts
acuiuli: This is the error response of the pdv tokenizer api. I don't know if we will use all of it but I mapped all of them in case the caller of the tokenizer would need it.

apps/etl-func/src/domain/interfaces/tokenizer.ts
acuiuli: this is a domain logic specific for the tokenization of sensible information of the message and I think it will be superfluous to create another file just for that. I think of this method as if it was a default method of the interface

apps/etl-func/local.settings.json.example
```diff
@@ -3,5 +3,7 @@
   "Values": {
     "AzureWebJobsStorage": "UseDevelopmentStorage=true",
     "FUNCTIONS_WORKER_RUNTIME": "node"
+    "PDV_TOKENIZER_API_KEY": "apiKey",
+    "PDV_TOKENIZER_BASE_URL": "https://pdv_base_url.it"
```
lucacavallaro: put uat link

apps/etl-func/src/adapters/env.ts
```diff
@@ -1,3 +1,8 @@
 import { z } from "zod";
 
-export const envSchema = z.object({}).and(z.record(z.string(), z.string()));
+export const envSchema = z
+  .object({
+    PDV_TOKENIZER_API_KEY: z.string().min(1),
+    PDV_TOKENIZER_BASE_URL: z.string().min(1),
```
lucacavallaro:
```suggestion
    PDV_TOKENIZER_BASE_URL: z.string().url(),
```

apps/etl-func/src/adapters/env.ts
```diff
@@ -1,3 +1,8 @@
...
+  })
+  .and(z.record(z.string(), z.string()));
```
lucacavallaro:
```suggestion
  ;
```

apps/etl-func/src/adapters/pdv-tokenizer/pdv-tokenizer-client.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export default class PDVTokenizerClient implements TokenizerClient {
+  #apiKey: string;
+  #baseUrl: string;
+
+  constructor(apiKey: string, baseUrl: string) {
+    assert.ok(apiKey, new Error("Api key is required"));
+    this.#apiKey = apiKey;
+    this.#baseUrl = baseUrl;
+  }
```
lucacavallaro:
```suggestion
  constructor(apiKey: string, baseUrl: string) {
    this.#apiKey = apiKey;
    this.#baseUrl = baseUrl;
  }
```

apps/etl-func/src/adapters/pdv-tokenizer/pdv-tokenizer-client.ts
```diff
@@ -0,0 +1,70 @@
...
+      } catch (error) {
+        attempt++;
+        if (!retryCondition(error) || attempt >= retries) {
+          throw error;
+        }
+        await delay(delayMs);
+      }
+    }
+    throw new Error(`Error during tokenizer api call after ${retries} retries`);
+  }
```
lucacavallaro: remove this

apps/etl-func/src/adapters/pdv-tokenizer/pdv-tokenizer-client.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  async maskSensitiveInfo(pii: string): Promise<string> {
```
lucacavallaro:
```suggestion
  async tokenize(pii: string): Promise<string> {
```

or `createToken` (`POST /tokens`)

apps/etl-func/src/adapters/pdv-tokenizer/pdv-tokenizer-client.ts
```diff
@@ -0,0 +1,70 @@
...
+        });
+
+        const responseJson = await response.json();
+
+        if (response.ok) {
+          return tokenResourceSchema.parse(responseJson).token;
+        }
+        throw new Error("Error during tokenizer api call", {
+          cause: problemSchema.parse(responseJson),
+        });
```
lucacavallaro:
```suggestion
      assert.strictEqual(
        response.ok,
        false,
        new Error("Error during tokenizer api call", {
          cause: problemSchema.parse(responseJson),
        }),
      );
      return tokenResourceSchema.parse(responseJson).token;
```

apps/etl-func/src/domain/interfaces/tokenizer.ts
```diff
@@ -0,0 +1,20 @@
...
+  maskSensitiveInfo(pii: string): Promise<string>;
+}
+
+const messageWithoutPIISchema = z.object({
+  fiscalCode: z.string().uuid().min(1),
+});
+
+export type MessageWithoutPII = z.infer<typeof messageWithoutPIISchema>;
+
+export const tokenize =
```
lucacavallaro:
```suggestion
export const maskSensitiveInfo =
```

apps/etl-func/src/domain/interfaces/tokenizer.ts
```diff
@@ -0,0 +1,20 @@
...
+  (message: { fiscalCode: string }) =>
```
lucacavallaro:
```suggestion
  (message: Message) =>
```

apps/etl-func/src/domain/interfaces/tokenizer.ts
```diff
@@ -0,0 +1,20 @@
+import { z } from "zod";
+
+export interface TokenizerClient {
+  maskSensitiveInfo(pii: string): Promise<string>;
+}
+
+const messageWithoutPIISchema = z.object({
```
lucacavallaro: idea: `messageSchema.omit({ fiscalCode: true }).extend(z.object({ recipientId: z.string().min(1) })`

lucacavallaro (changes requested):

apps/etl-func/src/domain/interfaces/tokenizer.ts
acuiuli: I had to use the MessageMetaData type instead

apps/etl-func/src/adapters/config.ts
```diff
@@ -7,6 +7,10 @@ export const configSchema = z.object({
     accountUri: z.string().url(),
     containerName: z.string().min(1),
   }),
+  pdvTokenizer: z.object({
+    apiKey: z.string().min(1),
+    baseUrl: z.string().min(1),
```
lucacavallaro:
```suggestion
    baseUrl: z.string().url(),
```

apps/etl-func/src/adapters/pdv-tokenizer/pdv-tokenizer-client.ts
```diff
@@ -0,0 +1,39 @@
...
+  async tokenize(pii: string): Promise<string> {
+    const request = piiResourceSchema.parse({ pii });
+    const response = await fetch(`${this.#baseUrl}/tokens`, {
+      body: JSON.stringify(request),
+      headers: {
+        "content-type": "application/json",
+        "x-api-key": this.#apiKey,
+      },
+      method: "PUT",
+    });
```
lucacavallaro:
```suggestion
    const response = await fetch(`${this.#baseUrl}/tokens`, {
      body: JSON.stringify({ pii }),
      headers: {
        "content-type": "application/json",
        "x-api-key": this.#apiKey,
      },
      method: "PUT",
    });
```

apps/etl-func/src/adapters/pdv-tokenizer/pii-resource.ts
lucacavallaro: no longer needed

lucacavallaro (changes requested):

apps/etl-func/src/domain/interfaces/tests/tokenizer.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const aFiscalCode = "AAADPZ44E08F367A";
+const aTokenizedFiscalCode = "3f5a5e37-63a0-423c-a108-94b535e03f91";
+
+const tokenizerClient: TokenizerClient = {
+  tokenize: async () => aTokenizedFiscalCode,
+};
+
+describe("tokenizeMessage", () => {
+  it("should return a message with the fiscal code tokenized if the tokenizer client api is successful", async () => {
```
Garma00:
```suggestion
  it("should return a tokenized fiscal code if the tokenizer client api is successful", async () => {
```

Garma00 (changes requested):

Garma00 (approved):

infra/resources/_modules/web_apps/etl_func.tf
```diff
@@ -17,12 +17,13 @@ module "etl_func" {
   health_check_path = "/api/health"
 
   app_settings = {
-    NODE_ENVIRONMENT               = "production",
+    NODE_ENVIRONMENT                    = "production",
```
lucacavallaro:
```suggestion
    NODE_ENV                    = "production",
```

lucacavallaro (approved):

PR #59, opened 2024-11-19
feat: add createMessageStatusEntity

Garma00:
#### List of Changes

* Add `createMessageStatusEntity`

#### Motivation and Context

We need a function that create the status entity from an unknown input that will be called with change feed elements.

Unit test

#### Types of changes

#### Checklist:


apps/etl-func/src/domain/message-status.ts
```diff
@@ -0,0 +1,37 @@
...
+]);
+
+const notRejectedMessageStatusSchema = z.object({
+  fiscalCode: z.string().min(1).optional(),
+  isArchived: z.boolean().default(false),
+  isRead: z.boolean().default(false),
+  messageId: z.string().min(1),
+  status: notRejectedMessageStatusValueEnum,
+  updatedAt: z.number(),
+});
```
acuiuli:
```suggestion
const statusEnum = z.enum([
  "ACCEPTED",
  "THROTTLED",
  "FAILED",
  "PROCESSED",
  "REJECTED",
]);

const rejectedMessageStatusSchema = z.object({
  fiscalCode: z.string().min(1).optional(),
  isArchived: z.boolean().default(false),
  isRead: z.boolean().default(false),
  messageId: z.string().min(1),
  rejection_reason: rejectionReasonSchema,
  status: statusEnum.extract(["REJECTED"]),
  updatedAt: z.number(),
});

const notRejectedMessageStatusSchema = z.object({
  fiscalCode: z.string().min(1).optional(),
  isArchived: z.boolean().default(false),
  isRead: z.boolean().default(false),
  messageId: z.string().min(1),
  status: statusEnum.exclude(["REJECTED"]),
  updatedAt: z.number(),
});
```

Why don't we have an enum that has all the status and then use zod to choose which one to use in the objects?

apps/etl-func/src/domain/message-status.ts
```diff
@@ -0,0 +1,37 @@
+import * as z from "zod";
+
+const rejectionReasonSchema = z
+  .enum(["SERVICE_NOT_ALLOWED", "USER_NOT_FOUND", "UNKNOWN"])
+  .default("UNKNOWN");
+
+const rejectedMessageStatusSchema = z.object({
+  fiscalCode: z.string().min(1).optional(),
```
acuiuli:
```suggestion
  fiscalCode: z.string().min(1),
```
Why don't we set the fiscalCode as mandatory? In the old schema it was optional because of backwards compatibility for older MessageStatus. We decided that the ingestion for the messageStatus will not have older items so I think it is better to have it mandatory.

apps/etl-func/src/domain/message-status.ts
Garma00: You are right, take a look here [45f9d40](https://github.com/pagopa/io-messages/pull/59/commits/45f9d40a83e65f22d3aa7f0864e59a9fab1238e1)

apps/etl-func/src/domain/message-status.ts
Garma00: You are right we could make it mandatory in this case

lucamancosu: can we use fiscalCode: fiscalCodeSchema instead ?

Garma00: [0530efe](https://github.com/pagopa/io-messages/pull/59/commits/0530efe8115fb8a44f61fe63cb658b2be506ad3d)

apps/etl-func/src/domain/message-status.ts
```diff
@@ -0,0 +1,49 @@
...
+  "FAILED",
+  "PROCESSED",
+  "REJECTED",
+]);
+
+const rejectedMessageStatusSchema = z.object({
+  fiscalCode: fiscalCodeSchema,
+  isArchived: z.boolean().default(false),
+  isRead: z.boolean().default(false),
+  messageId: z.string().min(1),
```
acuiuli: shouldn't this be an ulid?

acuiuli (approved):

PR #61, opened 2024-11-20
Iocom 1876 etl func compose

lucamancosu:
Basic composition of the etl_func procedures required for the message ingestion process.

#### List of Changes

The function is triggered by a CosmosDB trigger.
The function now have the roles needed to comunicate with the required data sources.
The function receive message from the CosmosDB, elaborate it and send it to the eventhub. 
For testing purpose only, i'm creating with terraform two containers inside the io-p-cosmos-api:
1. messages-ingestion-test
2. messages-ingestion-test-lease

#### Motivation and Context

This part aim to achieve the tests needed to for the approval of the data format sent to the Data Analytics team.

I've started the function locally.
I've used the remote data sources listed in the local.settings.example.
If the lease schema(messages-ingestion-lease) is emty than the function process all the messages.
I've tested only the 8+ messages available on DEV env.

#### Types of changes

#### Checklist:


apps/etl-func/src/adapters/__tests__/message-event.test.ts
```diff
@@ -26,19 +27,19 @@ const sendBatchMock = vi.fn(() => Promise.resolve());
 
 const eventHubProducerClient = new mocks.EventHubProducerClient();
 
-const messageEventAdapter = new EventHubEventProducer(
+const messageEventAdapter = new EventHubEventProducer<MessageEvent>(
   eventHubProducerClient,
   messageSchema,
 );
-
+const aSimpleMessageEventArrat: MessageEvent[] = [aSimpleMessageEvent];
```
acuiuli:
```suggestion
const aSimpleMessageEventArray: MessageEvent[] = [aSimpleMessageEvent];
```

apps/etl-func/src/adapters/__tests__/message-event.test.ts
```diff
@@ -26,19 +27,19 @@ const sendBatchMock = vi.fn(() => Promise.resolve());
...
 describe("publishMessageEvent", () => {
   beforeEach(() => {
     tryAddMock.mockRestore();
     sendBatchMock.mockRestore();
   });
   test("Given a valid message event it should resolve", async () => {
     await expect(
-      messageEventAdapter.publish(aSimpleMessageEvent),
+      messageEventAdapter.publish(aSimpleMessageEventArrat),
```
acuiuli:
```suggestion
      messageEventAdapter.publish(aSimpleMessageEventArray),
```

apps/etl-func/src/adapters/__tests__/message-event.test.ts
```diff
@@ -47,7 +48,7 @@ describe("publishMessageEvent", () => {
   test("Should throw an error if tryAddMock returns false", async () => {
     tryAddMock.mockImplementation(() => false);
     await expect(
-      messageEventAdapter.publish(aSimpleMessageEvent),
+      messageEventAdapter.publish(aSimpleMessageEventArrat),
```
acuiuli: ditto

apps/etl-func/src/adapters/__tests__/message-event.test.ts
```diff
@@ -57,7 +58,7 @@ describe("publishMessageEvent", () => {
   test("Should throw an error if sendBatchMock rejects", async () => {
     sendBatchMock.mockImplementation(() => Promise.reject());
     await expect(
-      messageEventAdapter.publish(aSimpleMessageEvent),
+      messageEventAdapter.publish(aSimpleMessageEventArrat),
```
acuiuli: ditto

apps/etl-func/src/adapters/config.ts
```diff
@@ -8,6 +8,11 @@ export const configSchema = z.object({
     accountUri: z.string().url(),
     containerName: z.string().min(1),
   }),
+  messageCosmosDB: z.object({
+    accountUri: z.string(),
```
acuiuli:
```suggestion
    accountUri: z.string().url(),
```

apps/etl-func/src/adapters/pdv-tokenizer/pdv-tokenizer-client.ts
```diff
@@ -44,13 +44,8 @@ export default class PDVTokenizerClient implements TokenizerClient {
 
       assert.strictEqual(
         response.ok,
-        false,
-        new Error(
-          `Error in tokenizer api call with status ${response.status}`,
-          {
-            cause: problemSchema.parse(responseJson),
-          },
```
acuiuli, line 52: why did you remove this? The cause contains the error response of the PDV api and it could be useful to debug errors

lucamancosu: I was having a problem running the pdv tokenization part. It seems that problemSchema.parse(responseJson) executes whether the assertion is successfull or not. So if the assertion is successfully(it means the tokenize api call goes well) it trhows an error because responseJson doesn't have the expected schema.

apps/etl-func/local.settings.json.example
```diff
@@ -8,6 +8,9 @@
     "EVENTHUB_CONNECTION_URI": "io-d-evh-cdc.servicebus.windows.net",
     "MESSAGE_EVENTHUB_NAME": "message-events-test"
     "PDV_TOKENIZER_API_KEY": "apiKey",
-    "PDV_TOKENIZER_BASE_URL": "https://api.uat.tokenizer.pdv.pagopa.it"
+    "PDV_TOKENIZER_BASE_URL": "https://api.uat.tokenizer.pdv.pagopa.it/tokenizer/v1",
+    "COSMOSDB_URI__accountEndpoint": "https://io-d-itn-common-cosno-01.documents.azure.com:443",
+    "COSMOSDB_DATABASE_NAME": "io-messages",
```
lucacavallaro: use `COSMOS` as "prefix" for all cosmos db related variables in order to be aligned to other resources like storage

(the `FunctionOptions`'s `connection` property requires a **prefix**, that can group several, related, variables)

```suggestion
    "COSMOS__accountEndpoint": "https://io-d-itn-common-cosno-01.documents.azure.com:443",
    "COSMOS_DBNAME": "io-messages",
```

apps/etl-func/local.settings.json.example
```diff
@@ -8,6 +8,9 @@
...
+    "MESSAGES_COSMOSDB_COLLECTION_NAME": "messages"
```
lucacavallaro: In Azure/CosmosDB [terminology](https://learn.microsoft.com/en-us/azure/cosmos-db/resource-model), `messages` is a `Container` not a `Collection` (unlike MongoDB and other DBMS) 

I also prefixed it with `COSMOS`

```suggestion
    "COSMOS_MESSAGES_CONTAINER_NAME": "messages"
```

apps/etl-func/src/adapters/config.ts
```diff
@@ -8,6 +8,11 @@ export const configSchema = z.object({
     accountUri: z.string().url(),
     containerName: z.string().min(1),
   }),
+  messageCosmosDB: z.object({
+    accountUri: z.string().url(),
+    containerName: z.string().min(1),
+    databaseName: z.string(),
+  }),
```
lucacavallaro:
```suggestion
  cosmos: z.object({
    accountUri: z.string().url(),
    databaseName: z.string().min(1),
    messagesContainerName: z.string().min(1).default("messages");
  }),
```

apps/etl-func/src/func.ts
```diff
@@ -34,6 +60,43 @@ const main = async (config: Config) => {
...
+    await producer.publish(messagesToSend);
+  };
+
+  app.cosmosDB("messagesCosmosDBTrigger", {
+    connection: "COSMOSDB_URI",
+    containerName: config.messageCosmosDB.containerName,
+    createLeaseContainerIfNotExists: true,
+    databaseName: config.messageCosmosDB.databaseName,
+    handler: messageIngestion,
+    leaseContainerName: `${config.messageCosmosDB.containerName}-ingestion-lease`,
```
lucacavallaro: is this required? maybe we can use the default name

apps/etl-func/src/func.ts
```diff
@@ -34,6 +60,43 @@ const main = async (config: Config) => {
...
+      }
+    }
+
+    await producer.publish(messagesToSend);
+  };
+
+  app.cosmosDB("messagesCosmosDBTrigger", {
+    connection: "COSMOSDB_URI",
+    containerName: config.messageCosmosDB.containerName,
+    createLeaseContainerIfNotExists: true,
```
lucacavallaro: ⚠️ we don't want to create containers outside terraform

```suggestion
    createLeaseContainerIfNotExists: false,
```

apps/etl-func/src/func.ts
```diff
@@ -34,6 +60,43 @@ const main = async (config: Config) => {
...
+        messagesToSend.push(messageEvent);
+      } else {
+        logger.info(
+          `${messageMetadata.id} returned an undefined message event`,
+        );
+      }
+    }
+
+    await producer.publish(messagesToSend);
+  };
```
lucacavallaro: can you move the handler and the function definition in a dedicated module?

apps/etl-func/src/func.ts
```diff
@@ -34,6 +60,43 @@ const main = async (config: Config) => {
     methods: ["GET"],
     route: "health",
   });
+
+  const messageIngestion = async (documents: unknown[]) => {
+    const messagesToSend: MessageEvent[] = [];
```
lucacavallaro: I think this array can be populated in a functional way, we should try to make all immutable!

lucacavallaro (dismissed):

apps/etl-func/src/adapters/pdv-tokenizer/pdv-tokenizer-client.ts
acuiuli: that's very strange, that part should only run if the error is thrown. If this is the case I think if will be better to have an if instead of the assertion in order to put the error response in the cause

apps/etl-func/src/adapters/env.ts
```diff
@@ -1,10 +1,13 @@
 import { z } from "zod";
 
 export const envSchema = z.object({
-  EVENTHUB_CONNECTION_URI: z.string().url(),
+  COSMOSDB_DATABASE_NAME: z.string().min(1),
+  COSMOSDB_URI__accountEndpoint: z.string().url(),
+  EVENTHUB_CONNECTION_URI: z.string(),
```
Garma00, line 7: I would use a specific name for this event hub because we are going to create another event hub for message-status.
```suggestion
  MESSAGE_EVENTHUB_CONNECTION_URI: z.string(),
```

apps/etl-func/src/func.ts
```diff
@@ -34,6 +60,43 @@ const main = async (config: Config) => {
...
+
+  app.cosmosDB("messagesCosmosDBTrigger", {
+    connection: "COSMOSDB_URI",
+    containerName: config.messageCosmosDB.containerName,
+    createLeaseContainerIfNotExists: true,
+    databaseName: config.messageCosmosDB.databaseName,
+    handler: messageIngestion,
+    leaseContainerName: `${config.messageCosmosDB.containerName}-ingestion-lease`,
+    maxItemsPerInvocation: 30,
+    startFromBeginning: true,
```
Garma00: If we want to use `startFromBeginning: true`, then we should provide a filter in the business logic in order to process only from the date that we want

Garma00 (changes requested):

apps/etl-func/src/func.ts
lucamancosu: the defualt name is **lease**.  There is a a high possibility that lease container is already used by other services in production than i've decided to divide them at least for this test scenario.

lucacavallaro: you should use `leaseContainerPrefix` instead (https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-cosmosdb-v2-trigger?tabs=python-v2%2Cisolated-process%2Cextensionv4%2Cnodejs-v4&pivots=programming-language-javascript#configuration)

setting a prefix enables us to reuse the `lease` container if it exists

lucamancosu: Yes we can try like that but in the io-p-comsos-api production dbs there is no **lease** named container and it seems that each service is creating is own one.

so what we can do is:

1. create a new one test ingestion lease like i'm doing. Creating a new one i think is more safe for testing purpose and we can delete it once the test is finished(i'm going to create it with terraform)
2. you can indicate to me which of the production lease container we must share with other services/functions and so we can use leaseContainerPrefix. In this last case we need to be able to delete the leases items for do a test retry if something goes wrong with the function execution (it can be a blocking and temporary error with PDV for example )and i think this is too risky in the prod enviroment with a shared container.

what do you think ?

PS: now i'm creating two containers inside the io-p-cosmos-api: **messages-ingestion-test**, **messages-ingestion-test-lease**

apps/etl-func/src/adapters/env.ts
lucamancosu: EVENTHUB_CONNECTION_URI is the eventhub namespace FQDN.
If we are going to use the same namespace, and i think this is probably the case for message-status, this remains the same and if you want we can call it EVENTHUB_NAMESPACE_CONNECTION_URI instead (but doesnt sound good to me).

If you mean that we are going to interact with multiple eventhub namespaces in the future than i agree with you and we must specialized this one like MESSAGES:EVENTHUB_CONNECTION_URI

what do you think ?

apps/etl-func/src/func.ts
lucamancosu: Yes that's right. I'm using only this property right now to avoid possible problems with the dataset prepared for the test. 

We can choose between:
1. add the required parameter **startFromTime** only after the test succeeded.
2. use for the tests **startFromTime** parameter

apps/etl-func/docs/dataplan-ingestion/etl-ingestion-flow.puml
Garma00, line 1: Cn you generate the png as well?

apps/etl-func/src/adapters/__tests__/message-event.test.ts
```diff
@@ -26,19 +27,19 @@ const sendBatchMock = vi.fn(() => Promise.resolve());
 
 const eventHubProducerClient = new mocks.EventHubProducerClient();
 
-const messageEventAdapter = new EventHubEventProducer(
+const messageEventAdapter = new EventHubEventProducer<MessageEvent>(
   eventHubProducerClient,
   messageSchema,
 );
-
+const aSimpleMessageEventArray: MessageEvent[] = [aSimpleMessageEvent];
```
Garma00: Whenever it's possible I would let typescript infer the type automatically, in this case the compiler should understand that `aSimpleMessageEventArray` is of type `MessageEvent`
```suggestion
const aSimpleMessageEventArray = [aSimpleMessageEvent];
```

apps/etl-func/src/adapters/env.ts
Garma00: You are right @lucamancosu

apps/etl-func/src/adapters/functions/__tests__/messages-ingestion.test.ts
```diff
@@ -0,0 +1,100 @@
...
+
+import messagesIngestion from "../messages-ingestion.js";
+
+const errorLogMock = vi.fn();
+const warnLogMock = vi.fn();
+
+const loggerMock = {
+  error: errorLogMock,
+  warn: warnLogMock,
+} as unknown as Logger;
```
Garma00: As we discussed before we could try to directly use the pino logger instead of mocking it. We can then spyOn the functions that are needed, I know that we followed this way before but I think that it is better to just use the spyOn from `vitest`.

apps/etl-func/src/adapters/functions/__tests__/messages-ingestion.test.ts
```diff
@@ -0,0 +1,100 @@
...
+  eventHubProducerClient,
+  messageSchema,
+);
+const publishSpy = vi.spyOn(producer, "publish").mockResolvedValue();
+
+const context = new InvocationContext();
+const handler = messagesIngestion(messageAdapter, PDVTokenizer, producer);
+
+describe("messagesIngestion handler", () => {
+  test("shoud resolve if nothing trows", async () => {
```
Garma00: I would modify this test case using an array with 2 elements so we can test that `getMessageByMetadataSpy` and `tokenizeSpy` are called `n` times. Also the publish spy should be called once.
We already have the test below to ensure that those functions are called once.

PS: there is a typo in the test description `trows`.

apps/etl-func/src/adapters/functions/messages-ingestion.ts
```diff
@@ -0,0 +1,57 @@
...
+    try {
+      //Retrieving the message contents for each message metadata
+      const messagesContet = (
+        await Promise.all(
+          documentsMetadata.map((messageMetadata) =>
+            messageAdapter.getMessageByMetadata(messageMetadata),
+          ),
+        )
+      ).filter(
+        (item: Message | undefined): item is Message => item !== undefined,
```
Garma00: Ditto, typescript can infer the type automatically here
```suggestion
        (item): item is Message => item !== undefined,
```

apps/etl-func/src/func.ts
Garma00: Yess, just remember this before going into production 😄

infra/resources/_modules/cosmos/cosmos_etl_messages.tf
```diff
@@ -0,0 +1,25 @@
...
+module "comsosdb_sql_container_messages_ingestion_test_lease" {
+  source = "github.com/pagopa/terraform-azurerm-v3//cosmosdb_sql_container?ref=v8.28.2"
+
+  name                = "messages-ingestion-test-lease"
+  resource_group_name = var.cosmosdb_account.resource_group_name
+
+  account_name  = var.cosmosdb_account.name
+  database_name = "db"
+
+  partition_key_path = "/id"
```
Garma00, line 23: The partition key for messages should be `fiscalCode` I think cc @lucacavallaro @acuiuli

Garma00 (changes requested):

infra/resources/_modules/cosmos/cosmos_etl_messages.tf
lucamancosu: Those 2 new containers are test containers only. This specific one is the lease test container and i don't really now if it knows fiscalCode as information. Instead the partition key on the same file refers to the messages-test-container that is the one we are going to populate with 100 test messages.

I think this 2 containers must be deleted anyway after the tests are closed because we are going to use the **messages** container instead and we are going to create a specific lease collection for the ingestion flow. 

@lucacavallaro @Garma00 @acuiuli  make sense ?

apps/etl-func/src/adapters/functions/__tests__/messages-ingestion.test.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  test("should not call any function if the documents array is empty", async () => {
+    const documentsMock: unknown[] = [];
+    await expect(handler(documentsMock, context)).resolves.toEqual(undefined);
+    expect(getMessageByMetadataSpy).not.toHaveBeenCalledOnce();
+    expect(tokenizeSpy).not.toHaveBeenCalledOnce();
+    expect(publishSpy).not.toHaveBeenCalledOnce();
+  });
+
+  test.only("should throw an error if tokenize throws an error", async () => {
```
Garma00:
```suggestion
  test("should throw an error if tokenize throws an error", async () => {
```

apps/etl-func/docs/messages-ingestion/messages-ingestion-flow.puml
```diff
@@ -0,0 +1,41 @@
...
+  etlFunc -> MessageContent: Get message content
+  break Error searching message content
+      etlFunc -> etlFunc: Throw error and exit
+  end
+  MessageContent -> etlFunc: Returns message content
+  note left of etlFunc
+    At this point the function knows
+    all the content of each message metadata
+  end note
+  etlFunc -> PDV: Request FC anonymization
```
Garma00:
```suggestion
  end note
  etlFunc -> etlFunc: Transform Message to MessageEvent
  etlFunc -> PDV: Request FC anonymization
```

apps/etl-func/src/adapters/config.ts
```diff
@@ -20,6 +25,11 @@ export type Config = z.TypeOf<typeof configSchema>;
 export const configFromEnvironment = envSchema
   .transform(
     (env): Config => ({
+      cosmos: {
+        accountUri: env.COSMOS__accountEndpoint,
+        databaseName: env.COSMOS_DBNAME,
+        messagesContainerName: env.COSMOS_MESSAGES_CONTAINER_NAME,
+      },
```
Garma00, line 32: we should add an env variable for `startFromTime`

apps/etl-func/src/func.ts
```diff
@@ -34,6 +57,27 @@ const main = async (config: Config) => {
...
     route: "health",
   });
+
+  app.cosmosDB("messagesCosmosDBTrigger", {
+    connection: "COSMOS",
+    containerName: config.cosmos.messagesContainerName,
+    createLeaseCollectionIfNotExists: false,
+    databaseName: config.cosmos.databaseName,
+    handler: messagesIngestion(messageAdapter, PDVTokenizer, producer),
+    leaseContainerName: `${config.cosmos.messagesContainerName}-ingestion-lease`,
```
Garma00: We should hardcode the name here in my opinion, if the container name changes we do not want to re-ingest.

infra/resources/_modules/cosmos/cosmos_etl_messages.tf
```diff
@@ -0,0 +1,25 @@
+module "comsosdb_sql_container_messages_ingestion_test" {
+  source = "github.com/pagopa/terraform-azurerm-v3//cosmosdb_sql_container?ref=v8.28.2"
+
+  name                = "messages-ingestion-test"
+  resource_group_name = var.cosmosdb_account.resource_group_name
+
+  account_name  = var.cosmosdb_account.name
+  database_name = "db"
+
+  partition_key_path = "/id"
```
Garma00: We should use `fiscalCode` as partition key here.

infra/resources/_modules/cosmos/cosmos_etl_messages.tf
```diff
@@ -0,0 +1,25 @@
...
+  default_ttl        = "-1"
+}
+
+module "comsosdb_sql_container_messages_ingestion_test_lease" {
+  source = "github.com/pagopa/terraform-azurerm-v3//cosmosdb_sql_container?ref=v8.28.2"
+
+  name                = "messages-ingestion-test-lease"
```
Garma00: This name should match the one that we are passing into the config.

apps/etl-func/docs/messages-ingestion/messages-ingestion-flow.puml
```diff
@@ -10,18 +10,19 @@ CosmosDB -> etlFunc: Messages metadata from CosmosDB
 
 loop Process all Mesagges Metadata
   etlFunc -> MessageContent: Get message content
-  break Error searching message content
-      etlFunc -> etlFunc: Throw error and exit
-  end
+  ' break Error searching message content
+  '     etlFunc -> etlFunc: Throw error and exit
+  ' end
```
Garma00: we can remove those

apps/etl-func/docs/messages-ingestion/messages-ingestion-flow.puml
```diff
@@ -10,18 +10,19 @@ CosmosDB -> etlFunc: Messages metadata from CosmosDB
...
     all the content of each message metadata
   end note
+  etlFunc -> etlFunc: Transform message to messageEvent
   etlFunc -> PDV: Request FC anonymization
-  break Anonymization error
-      etlFunc -> etlFunc: Throw error and exit
-  end
+  ' break Anonymization error
+  '     etlFunc -> etlFunc: Throw error and exit
+  ' end
```
Garma00: ditto

Garma00 (changes requested):

Garma00 (approved):

acuiuli (approved):

PR #63, opened 2024-11-21
feat: Setting up load message status

Garma00:
#### List of Changes

* Add new event hub to infra;
* Change configuration to use the new event hub as well;
* Update healthcheck function.

#### Motivation and Context

We need a method to load message status into the event hub.

#### Types of changes

#### Checklist:


infra/resources/_modules/web_apps/etl_func.tf
```diff
@@ -17,13 +17,15 @@ module "etl_func" {
...
-    MESSAGE_CONTENT_STORAGE_URI    = var.app_settings.message_content_storage_uri
-    EVENTHUB_CONNECTION_URI        = var.app_settings.eventhub_connection_uri,
-    MESSAGE_CONTENT_CONTAINER_NAME = "message-content",
-    MESSAGE_EVENTHUB_NAME          = "io-p-itn-com-etl-messages-evh-01"
-    PDV_TOKENIZER_API_KEY = "@Microsoft.KeyVault(VaultName=${var.common_key_vault.name};SecretName=func-elt-PDV-TOKENIZER-API-KEY)"
+    NODE_ENV                               = "production",
+    FUNCTIONS_WORKER_RUNTIME               = "node",
+    MESSAGE_CONTENT_STORAGE_URI            = var.app_settings.message_content_storage_uri
+    MESSAGE_EVENTHUB_CONNECTION_URI        = var.app_settings.eventhub_connection_uri,
+    MESSAGE_STATUS_EVENTHUB_CONNECTION_URI = var.app_settings.eventhub_connection_uri,
```
lucacavallaro: if the value is the same, why we have two different variables?

lucacavallaro (changes requested):

infra/resources/_modules/web_apps/etl_func.tf
Garma00: It can be useful to use two different variables here because we could use two different event hub easily (for example for testing purpose). If you think that it is better to use just one var I can refactor it.

lucacavallaro: now we are using only one event hub namespace, so I would declare only one variable!

Garma00: Take a look here [9d5999e](https://github.com/pagopa/io-messages/pull/63/commits/9d5999ebda97774a38f0d76fb3b371c8fec7a07c)

apps/etl-func/local.settings.json.example
```diff
@@ -5,8 +5,9 @@
     "FUNCTIONS_WORKER_RUNTIME": "node",
     "MESSAGE_CONTENT_STORAGE_URI": "https://testgiovannist.blob.core.windows.net/",
     "MESSAGE_CONTENT_CONTAINER_NAME": "messages",
-    "EVENTHUB_CONNECTION_URI": "io-d-evh-cdc.servicebus.windows.net",
+    "MESSAGE_EVENTHUB_CONNECTION_URI": "io-d-evh-cdc.servicebus.windows.net",
```
lucamancosu: If we want to change this variable name than those files should be updated:


apps/etl-func/src/adapters/config.ts
apps/etl-func/src/adapters/env.ts
infra/resources/_modules/web_apps/locals.tf
infra/resources/_modules/web_apps/variables.tf
infra/resources/prod/web_apps.tf

apps/etl-func/local.settings.json.example
```diff
@@ -5,8 +5,9 @@
...
     "MESSAGE_EVENTHUB_NAME": "message-events-test"
+    "MESSAGE_STATUS_EVENTHUB_NAME": "message-status-events-test"
```
lucamancosu: ditto

apps/etl-func/local.settings.json.example
Garma00: We don't need to update the name of this variable because we are going to use the same event hub, ita was a mistake during the merge I believe, nice catch.
[2e3aa71](https://github.com/pagopa/io-messages/pull/63/commits/2e3aa712168c5b40970a4301115db7107e85a896)

lucamancosu (approved):

lucacavallaro (approved):

PR #66, opened 2024-11-22
feat(etl-func): add redis cache to tokenizer

acuiuli:
#### List of Changes

- [feat(etl-func): add redis cache to tokenizer](https://github.com/pagopa/io-messages/pull/66/commits/e54070f0e6f306c21f571628655f0e8b282072d6)

#### Motivation and Context

This pr adds a redis cache to the tokenizer in order to avoid calling the PDV tokenize api multiple times for the same fiscal code and improve performance

unit test

#### Types of changes

#### Checklist:


lucacavallaro:
PR description needed

apps/etl-func/src/adapters/redis/config.ts
```diff
@@ -0,0 +1,9 @@
+import { z } from "zod";
+
+export const redisConfigSchema = z.object({
+  accessKey: z.string().min(1),
+  pingInterval: z.number(),
```
Garma00: Can we set a min here? I would suggest 1 second here
```suggestion
  pingInterval: z.number().min(1000),
```

infra/resources/_modules/web_apps/etl_func.tf
```diff
@@ -23,7 +23,10 @@ module "etl_func" {
     EVENTHUB_CONNECTION_URI        = var.app_settings.eventhub_connection_uri,
     MESSAGE_CONTENT_CONTAINER_NAME = "message-content",
     MESSAGE_EVENTHUB_NAME          = "io-p-itn-com-etl-messages-evh-01"
-    PDV_TOKENIZER_API_KEY = "@Microsoft.KeyVault(VaultName=${var.common_key_vault.name};SecretName=func-elt-PDV-TOKENIZER-API-KEY)"
+    PDV_TOKENIZER_API_KEY          = "@Microsoft.KeyVault(VaultName=${var.common_key_vault.name};SecretName=func-elt-PDV-TOKENIZER-API-KEY)"
+    REDIS_URL                      = var.redis_cache.url
+    REDIS_PASSWORD                 = var.redis_cache.access_key
+    REDIS_PING_INTERVAL            = 1000 * 60 * 9
```
Garma00: Isn't this value too high?

apps/etl-func/src/adapters/__tests__/message.test.ts
```diff
@@ -31,10 +36,25 @@ const messageContentMock: MessageContentProvider = {
...
-const tokenizeMock = vi.fn();
-const tokenizerClientMock = {
-  tokenize: tokenizeMock,
-} as unknown as PDVTokenizerClient;
+const recipientStore = new Map<string, unknown>();
+
+const tokenizerClient: Mocked<TokenizerClient> = {
+  tokenize: vi
+    .fn()
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
Garma00, line 40: is this comment needed?

Garma00 (changes requested):

apps/etl-func/src/adapters/__tests__/message.test.ts
acuiuli: if i remove it the linter will fail because the `_fiscalCode` variable is not used

infra/resources/_modules/web_apps/etl_func.tf
acuiuli: In the azure documentation it said that it should be lees than 10 minutes because after 10 minute redis will clean up the connection, so I set it for 9 minutes. https://learn.microsoft.com/en-us/azure/azure-cache-for-redis/cache-best-practices-connection#idle-timeout

Garma00 (approved):

apps/etl-func/src/domain/interfaces/__tests__/tokenizer.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+    recipientStore.set(`user:${fiscalCode.toLowerCase()}:recipient.id`, event);
+  }),
+};
+
+describe("tokenizeMessage", () => {
+  it("should return a tokenized fiscal code if the tokenizer client api is successful", async () => {
+    const run = await tokenize(aFiscalCode)({
+      recipientRepository,
+      tokenizerClient,
+    } as TokenizerEnvironment);
```
lucacavallaro: why is this type casting needed?

apps/etl-func/src/domain/interfaces/tokenizer.ts
```diff
@@ -1,8 +1,31 @@
+import { RecipientIdNotFoundError } from "@/adapters/redis/recipient.js";
+
 export interface TokenizerClient {
   tokenize(pii: string): Promise<string>;
 }
+export interface RecipientRepository {
+  get(fiscalCode: string): Promise<null | string>;
```
lucacavallaro: why `null` and not `undefined`?

apps/etl-func/src/domain/mask-fiscal-code.ts
```diff
@@ -0,0 +1,27 @@
...
+} from "./interfaces/tokenizer.js";
+
+export class MaskFiscalCodeUseCase {
+  #ctx: TokenizerEnvironment;
+
+  constructor(ctx: TokenizerEnvironment) {
+    this.#ctx = ctx;
+  }
+
+  async execute(fiscalCode: string): Promise<string> {
```
lucacavallaro: maybe instead of a string you can use the `FiscalCode` type (that is a branded type)

apps/etl-func/src/domain/interfaces/tokenizer.ts
```diff
@@ -1,8 +1,31 @@
...
+export interface TokenizerEnvironment {
+  recipientRepository: RecipientRepository;
+  tokenizerClient: TokenizerClient;
+}
+
+export const getCachedRecipientId =
+  (fiscalCode: string) =>
+  async ({
+    recipientRepository: repo,
+  }: TokenizerEnvironment): Promise<string> => {
```
lucacavallaro: imho the caching its just a implementation detail, it should not be exposed on the domain layer

lucacavallaro: also I suggest you avoid writing this kind of "partial-applicable" function, since we are no longer enforcing a functional style

you can simply define `(repo, fiscalCode) => Promise<string>`

apps/etl-func/src/domain/interfaces/tokenizer.ts
```diff
@@ -1,8 +1,31 @@
...
+    const recipientId = await repo.get(fiscalCode);
+    if (!recipientId) {
+      throw new RecipientIdNotFoundError();
+    }
```
lucacavallaro: I think we should return `undefined` here, to make a clear distinction between a valid result ("the value does not exists on cache") and an error (redis is down)

apps/etl-func/src/domain/mask-fiscal-code.ts
```diff
@@ -0,0 +1,27 @@
...
+    try {
+      const recipientId = await getCachedRecipientId(fiscalCode)(this.#ctx);
+      return recipientId;
+    } catch (error) {
+      if (error instanceof RecipientIdNotFoundError) {
+        return await tokenize(fiscalCode)(this.#ctx);
+      }
+      throw error;
+    }
```
lucacavallaro: I think this whole logic should not be declared in a `use-case`, the application does not need to know that the tokenizer client can be cached

---

You have a `TokenizerClient` interface, that is implemented by the `PDVTokenizerClient` class, right? 

If so, you can just write a wrapper, `CachedPDVTokenizerClient` that satisfies the same interface  and use it without touching domain and application layer.

lucacavallaro (changes requested):

apps/etl-func/src/domain/mask-fiscal-code.ts
acuiuli: I created a CachedPDVTokenizerClient that extends the PDVTokenizerClient and adds the cache functionality

apps/etl-func/src/domain/interfaces/tokenizer.ts
acuiuli: I used null because the get method of the redis sdk returns null and not undefined. I modified it to return an undefined

apps/etl-func/src/domain/interfaces/__tests__/tokenizer.spec.ts
acuiuli: it is not needed, i removed it

infra/resources/_modules/web_apps/etl_func.tf
```diff
@@ -23,7 +23,10 @@ module "etl_func" {
     EVENTHUB_CONNECTION_URI        = var.app_settings.eventhub_connection_uri,
     MESSAGE_CONTENT_CONTAINER_NAME = "message-content",
     MESSAGE_EVENTHUB_NAME          = "io-p-itn-com-etl-messages-evh-01"
-    PDV_TOKENIZER_API_KEY = "@Microsoft.KeyVault(VaultName=${var.common_key_vault.name};SecretName=func-elt-PDV-TOKENIZER-API-KEY)"
+    PDV_TOKENIZER_API_KEY          = "@Microsoft.KeyVault(VaultName=${var.common_key_vault.name};SecretName=func-elt-PDV-TOKENIZER-API-KEY)"
+    REDIS_URL                      = var.redis_cache.url
+    REDIS_PASSWORD                 = var.redis_cache.access_key
```
lucamancosu: Probably this must be REDIS_ACCESS_KEY ?

apps/etl-func/local.settings.json.example
```diff
@@ -9,5 +9,8 @@
     "MESSAGE_EVENTHUB_NAME": "message-events-test"
     "PDV_TOKENIZER_API_KEY": "apiKey",
     "PDV_TOKENIZER_BASE_URL": "https://api.uat.tokenizer.pdv.pagopa.it"
+    "REDIS_ACCESS_KEY": "accessKey",
+    "REDIS_PING_INTERVAL": 1000
```
lucamancosu: some commas missing in the file. Maybe we can have problems after a copy for a local execution.

lucamancosu: I cant run the function locally passing "REDIS_PING_INTERVAL": 1000.

I get REDIS_PING_INTERVAL Expected number, received string.

acuiuli: It should be fixed [here](https://github.com/pagopa/io-messages/pull/66/commits/8629b029f7523fd1c2797f1798f230e2cc548a03)

infra/resources/_modules/web_apps/etl_func.tf
acuiuli: yes you are right

apps/etl-func/src/domain/interfaces/tokenizer.ts
```diff
@@ -1,8 +1,9 @@
...
 }
-
-export const maskSensitiveInfo =
-  (fiscalCode: string) =>
-  async (client: TokenizerClient): Promise<string> =>
-    await client.tokenize(fiscalCode);
+export interface RecipientRepository {
+  get(fiscalCode: FiscalCode): Promise<string | undefined>;
+  upsert(fiscalCode: FiscalCode, recipientId: string): Promise<void>;
+}
```
lucacavallaro: I don't understand why you created a `RecipientRepository` - wasn't `maskSensitiveInfo` the right abstraction?

What is the purpose of this repository?

acuiuli: this repository is the one responsible for the cached recipientId, I use it to check if a fiscalCode is already cached or not and to save the tokenized fiscalCode in the cache [here](https://github.com/pagopa/io-messages/pull/66/files#diff-e05293644adfa971f457c46d7f8998a3769bcd4dde3aae32e8359b548b2b338f:~:text=const%20recipientId%20%3D%20await%20this.%23recipientRepo.get(fiscalCode)%3B). This repository has the specific implementation for redis. 

The maskSensitiveInfo method I think was an over-engineering since it was just one line of code and didn't do anything different that the client didn't do already.

lucacavallaro: the fact that a certain fiscalCode is cached or not is just an implementation detail, and should not be important (and known) at the domain level

the domain just `tokenize` the fiscalCode, how is tokenized (by PDV, cached or not) is relevant only to infrastructure layer

acuiuli: ok I agree that the domain should not know if there is a caching or not when tokenizing a fiscalCode. I removed the recipientRepository interface [here](https://github.com/pagopa/io-messages/pull/66/commits/3ebb021a2e68623eb8e138959de5d14dd8b4a7c0).

Garma00 (approved):

lucacavallaro (approved):

PR #73, opened 2024-12-06
fix: prepare etl-func for ingestion

Garma00:
#### List of Changes

* Add a check to ignore all pending messages
* Increase batch elements to 50

Unit test

#### Types of changes

#### Checklist:


apps/etl-func/src/adapters/avro.ts
```diff
@@ -31,11 +31,6 @@ export const messageSchema = avro.Type.forSchema({
       name: "sender_user_id",
       type: "string",
     },
-    {
```
lucamancosu, line 34: We should change the apps/etl-func/avro/message.avsc too.

Should we remove it from apps/etl-func/src/__mocks__/message-event.ts ?

acuiuli (approved):

lucamancosu (approved):

PR #77, opened 2024-12-17
feat: add message-status-ingestion function

Garma00:
#### List of Changes

1. Add message-status-ingestion sequence diagram
2. Add new function for message status ingestion
3. Small fixes in some schemas

#### Motivation and Context

We need a new azure function for the message-status ingestion but we don't have to start it yet.

The function has been tested locally and the pull request also provides unit tests.

##### Sequence diagram
image

#### Types of changes

#### Checklist:


apps/etl-func/docs/message-status-ingestion/message-status-ingestion.puml
```diff
@@ -0,0 +1,25 @@
+@startuml message-status-ingestion
+
+collections "message-status-collection" as msc
+participant MessageStatusIngestion as msi
+collections "message-status-evh" as evh
+
+msc <- msi : Read batch of 50 documents using change feed trigger
```
lucamancosu: I suggest to invert the arrow direction. From my understanding is the resources, in this case the message-status-collection that sends the data to the function that is listening to the change feed. If you agree with me than i suggest to change the label too.

apps/etl-func/local.settings.json.example
```diff
@@ -10,11 +10,12 @@
     "MESSAGE_STATUS_EVENTHUB_NAME": "message-status-events-test",
     "PDV_TOKENIZER_API_KEY": "apiKey",
     "PDV_TOKENIZER_BASE_URL": "https://api.uat.tokenizer.pdv.pagopa.it/tokenizer/v1",
-    "REDIS_ACCESS_KEY": "accessKey",
+    "REDIS_PASSWORD": "accessKey",
     "REDIS_PING_INTERVAL": 1000,
     "REDIS_URL": "https://a-redis-url/"
```
lucamancosu: missing comma

apps/etl-func/src/adapters/avro.ts
```diff
@@ -126,3 +126,69 @@ export const messageSchema = avro.Type.forSchema({
...
+      type: "long",
+    },
+    {
+      doc: "A timestamp that indicates when the status was created.",
+      name: "created_at",
+      type: "long",
+    },
+  ],
+  name: "message_status",
+  namespace: "dto",
```
lucamancosu, line 192: should we need to change the namespace into it.ioapp.com.message ?

infra/resources/_modules/web_apps/locals.tf
```diff
@@ -4,21 +4,22 @@ locals {
...
+      MESSAGE_STATUS_EVENTHUB_NAME         = "io-p-itn-com-etl-message-status-evh-01"
+      PDV_TOKENIZER_API_KEY                = "@Microsoft.KeyVault(VaultName=${var.common_key_vault.name};SecretName=func-elt-PDV-TOKENIZER-API-KEY)"
+      PDV_TOKENIZER_BASE_URL               = "https://api.tokenizer.pdv.pagopa.it/tokenizer/v1"
+      REDIS_PASSWORD                       = var.redis_cache.access_key
+      REDIS_PING_INTERVAL                  = 1000 * 60 * 9
+      REDIS_URL                            = var.redis_cache.url
+      COSMOS__accountEndpoint              = var.cosmosdb_account_api.endpoint
+      COSMOS_DBNAME                        = "db",
+      COSMOS_MESSAGES_CONTAINER_NAME       = "messages-dataplan-ingestion-test"
+      COSMOS_MESSAGE_STATUS_CONTAINER_NAME = "message-status"
```
lucamancosu, line 22: if we need to test the function in production i think is better to create a message-status-ingestion-test collection. What you think ?

Garma00: I would do this in a separate task when we'll decide to do a test

apps/etl-func/src/adapters/avro.ts
Garma00: I would use dto because for this schema, we are required to use the same as the one used in https://github.com/pagopa/io-functions-elt

apps/etl-func/local.settings.json.example
Garma00: [a4ba27b](https://github.com/pagopa/io-messages/pull/77/commits/a4ba27b7d606fd498665b43d9048b230a8658897)

apps/etl-func/docs/message-status-ingestion/message-status-ingestion.puml
Garma00: [3e2bece](https://github.com/pagopa/io-messages/pull/77/commits/3e2bece0c846f960497a9c3d9078150c714d0e89), [28cdfea](https://github.com/pagopa/io-messages/pull/77/commits/28cdfeaeb046b879f453254b102752b79357611f)

lucamancosu (approved):

apps/etl-func/src/adapters/functions/message-status-ingestion.ts
```diff
@@ -0,0 +1,22 @@
...
+  return success ? data : undefined;
+};
+
+const messageStatusIngestionHandler =
+  (ingestUseCase: IngestMessageStatusUseCase): CosmosDBHandler =>
+  async (documents: unknown[]) => {
+    const parsedMessageStatusBatch = documents.map(parseMessageStatus);
+    const messageStatusBatch = parsedMessageStatusBatch.filter(
+      (status): status is MessageStatus => status !== undefined,
+    );
```
acuiuli: Can't we remove the parseMessageStatus function and connect the filter with the map to have it like this?
```suggestion
    const messageStatusBatch = documents
  .map((document) => {
    const { data, success } = messageStatusSchema.safeParse(document);
    return success ? data : undefined;
  })
  .filter((status): status is MessageStatus => status !== undefined);

```

apps/etl-func/src/domain/message-status.ts
```diff
@@ -17,12 +17,12 @@ export const messageStatusIdSchema = z.string().min(1);
 
 export const messageStatusSchema = z
   .object({
-    fiscalCode: fiscalCodeSchema,
+    fiscalCode: fiscalCodeSchema.optional(),
     id: messageStatusIdSchema,
```
acuiuli, line 26: is this messageStatusIdSchema really necessary? It is just a string

apps/etl-func/src/func.ts
```diff
@@ -107,6 +121,30 @@ const main = async (config: Config) => {
     //we need to start the ingestion from this date
     startFromTime: "2023/01/01T00:00:00Z",
   });
+
+  // NOTE: we don't want to start the ingestion yet
```
acuiuli, line 125: if we use the test container I think we can uncomment this right? In this way we can also do a test run if we want

apps/etl-func/src/domain/message-status.ts
Garma00: It is exported because we use it also in messageStatusEventSchema, this way if we modify it here we don't have to modify it in the event

acuiuli: Why don't we create the correct schema then? Something like this 
`const versionedId = z.string().regex(
  /^[0-9A-HJKMNP-TV-Z]{26}-\d+$/,
  'Invalid format. Expected <ULID-Number>'
);`

acuiuli (approved):

PR #79, opened 2024-12-19
Message ingestion - message metadata parse fix

lucamancosu:
Fixed message metadata parsing
The function can run with a redis local instance.

#### List of Changes

Added the missing parsing of message metadata.
Added the instructions to start a local containerized redis instance. We can now test the all function locally. 

I ran the function locally and completed the test ingestion flow

#### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

#### Checklist:


apps/etl-func/src/adapters/functions/messages-ingestion.ts
```diff
@@ -11,16 +11,22 @@ const logger = pino({
...
   (ingestUseCase: IngestMessageUseCase): CosmosDBHandler =>
   async (documents: unknown[]) => {
     //Avoiding all documents different from MessageMetadata schema and with
-    //isPending equals to true
-    const messagesMetadata: MessageMetadata[] = documents.filter(
-      (item): item is MessageMetadata => {
-        const parsedItem = messageMetadataSchema.safeParse(item);
-        return parsedItem.success && !parsedItem.data.isPending;
-      },
+    //isPending equals to trueù
```
Garma00:
```suggestion
    //isPending equals to true
```

Garma00 (approved):

apps/etl-func/src/adapters/functions/messages-ingestion.ts
```diff
@@ -11,16 +11,22 @@ const logger = pino({
...
     //isPending equals to true
-    const messagesMetadata: MessageMetadata[] = documents.filter(
-      (item): item is MessageMetadata => {
-        const parsedItem = messageMetadataSchema.safeParse(item);
-        return parsedItem.success && !parsedItem.data.isPending;
-      },
+
+    const parsedMessagesMetadata = documents.map(processMessageMetadata);
+    const messagesMetadata: MessageMetadata[] = parsedMessagesMetadata.filter(
+      (item): item is MessageMetadata => item !== undefined,
```
acuiuli, line 31: Can't we unify the code like this by removing the function and linking the map and the filter together?
```suggestion
const messagesMetadata: MessageMetadata[] = documents
  .map((input) => {
    const { data, success } = messageMetadataSchema.safeParse(input);
    return success && !data?.isPending ? data : undefined;
  })
  .filter((item): item is MessageMetadata => item !== undefined);
```

lucamancosu: I think in this way we are going to loose readability. Do you see any problem if we keep the current version ?

lucacavallaro: imho the cleanest solution is using `zod` to perform this validation

something like

```ts
const messagesMetadata = z.array(messageMetadataSchema).parse(documents)
```

(maybe, if needed, combined with something like zod's `.refine`)

lucacavallaro (dismissed):

apps/etl-func/src/adapters/functions/messages-ingestion.ts
Garma00: @lucacavallaro this way, we are unable to distinguish valid elements within the array. Our goal is to ignore invalid elements, but using this type of validation will result in the entire array being treated as invalid.

lucamancosu: f4fcffeb59c1b83bd3107b693ad586d5be2b572c

Garma00 (approved):

apps/etl-func/src/adapters/functions/messages-ingestion.ts
```diff
@@ -11,16 +11,24 @@ const logger = pino({
   level: process.env.NODE_ENV === "production" ? "error" : "debug",
 });
 
+const processMessageMetadata = (
+  input: unknown,
+): MessageMetadata | undefined => {
+  const result = messageMetadataSchema.safeParse(input);
+  //we are considering, and returning, only processed messages. A processed message has isPending === false
+  if (result.success && !result.data.isPending) return result.data;
+  else return undefined;
```
lucacavallaro, line 20: maybe it can be simplified in this way (?)

```suggestion
//we are considering, and returning, only processed messages. A processed message has isPending === false
const result = messageMetadataSchema
  .and(z.object({ isPending: z.literal(false) }))
  .catch(undefined)  
  .parse(input)
return result;
```

lucamancosu: I've tested it but there are types problems with the catch. I've decided to keep the if statement

PR #80, opened 2025-01-02
Improve toolchain

lucacavallaro:
## List of changes
1. Remove `integration-tests` from workspaces
2. Upgrade `turbo` to the latest major version (`2.x`)
3. Upgrade `node` to `20.18.1` (latest 20 LTS)
4. Upgrade `yarn` to `4.6.0` via `corepack`
5. Remove unused dependencies and config files from root
6. Remove `jest` from legacy apps (switched to `vitest`)
7. Enable [Yarn Plug'n'Play](https://yarnpkg.com/features/pnp)


apps/citizen-func/package.json
```diff
@@ -1,61 +1,45 @@
...
   "name": "citizen-func",
   "version": "5.5.4",
-  "license": "MIT",
+  "license": "UNLICENSED",
   "scripts": {
-    "prebuild": "yarn openapi:bundle && rimraf ./generated && yarn generate:definitions",
-    "build": "tsc",
-    "test": "jest",
-    "test:coverage": "jest --coverage",
+    "build": "yarn openapi:bundle && rimraf ./generated && yarn generate:definitions && tsc",
```
Garma00, line 6: @lucacavallaro I understand that this is intended to maintain the same behavior, but I would prefer to separate the `generate` and `build` steps. I believe that running the generate script every time you want to build slows down the development process. What do you think?

apps/sending-func/package.json
```diff
@@ -1,72 +1,51 @@
 {
   "name": "sending-func",
   "version": "1.10.12",
-  "license": "MIT",
+  "license": "UNLICENSED",
   "scripts": {
-    "prebuild": "yarn openapi:bundle && yarn generate:definitions && yarn generate:backend-notification-hub",
-    "build": "tsc",
+    "build": "yarn openapi:bundle && yarn generate:definitions && yarn generate:backend-notification-hub && tsc",
```
Garma00, line 6: ditto

Garma00 (approved):

apps/citizen-func/package.json
lucacavallaro: @Garma00 you right, but the goal of this change is to update the toolchain without changing its behavior. We will refactor it in the future, removing unnecessary steps and using better tools (for example `esbuild` instead of `tsc` for transpilation).

apps/sending-func/package.json
lucacavallaro: see https://github.com/pagopa/io-messages/pull/80#discussion_r1907202178

apps/citizen-func/package.json
lucacavallaro: (`yarn@4` does not support lifecycle scripts, so I had to merge `prebuild` and `build`)

acuiuli (approved):

PR #85, opened 2025-01-13
feat: add utilities to check message events based on time ranges

Garma00:
#### List of Changes

- Add a new cosmos event summary collector
- Launch a collect method every time the message ingestion publish events into the event hub
- Add unit test for the new cosmos event summary collector
- Fix message ingestion unit test after the collect method was added into the use case
- Update documentation

##### Cosmos event summary collector

It is a utility to store a `summary` into a cosmos container.
A summary is formed as follow:
```json
{
  "year": "2025"
  "id": "2025-W01"
  "count": 100
}
```
Where the count represent the number of events that were sent.
When the `collect` method is called it will:
1. calculate the `partitionKey` (the curent year)
2. calculate the `modelId` (the current year + the current week number)
3. Get the summary and update the count if a summary for the week already exists
4. Create a new summary if no summary exists

#### Motivation and Context

We need an efficient way to retrieve metrics related to the message ingestion such as:

- Number of events processed between 2 timestamp
- Number of events processed in a month or a year

Unit test.
Also the entire flow was tested locally on my machine.

#### Types of changes

#### Checklist:


apps/etl-func/local.settings.json.example
```diff
@@ -16,6 +16,9 @@
     "COSMOS__accountEndpoint": "https://io-d-itn-common-cosno-01.documents.azure.com:443",
     "COSMOS_DBNAME": "io-messages",
     "COSMOS_MESSAGES_CONTAINER_NAME": "messages",
+    "IOCOM_COSMOS__accountEndpoint": "https://io-d-itn-common-cosno-01.documents.azure.com:443",
```
lucamancosu, line 19: this seams a duplicate of the COSMOS__accountEndpoint. I propose to delete it or if we want to take IOCOM_COSMOS instead of COSMOS than it must be replaced everywhere.

apps/etl-func/local.settings.json.example
```diff
@@ -16,6 +16,9 @@
...
+    "IOCOM_COSMOS_EVENTS_COLLECTOR_DBNAME": "io-messages",
```
lucamancosu, line 20: is it a COSMOS_DBNAME duplicate ?

apps/etl-func/src/adapters/cosmos/config.ts
```diff
@@ -0,0 +1,9 @@
+import { z } from "zod";
+
+export const cosmosConfigSchema = z.object({
+  accountUri: z.string().regex(/^[a-zA-Z0-9-]+\.servicebus\.windows\.net$/),
```
lucamancosu: this regex is for a service bus

apps/etl-func/src/domain/use-cases/ingest-message.ts
```diff
@@ -38,5 +47,14 @@ export class IngestMessageUseCase {
       ),
     );
     if (messagesEvent.length) await this.#eventProducer.publish(messagesEvent);
+
+    try {
+      this.#eventSummaryCollector.collect(messagesEvent);
+    } catch (error) {
+      logger.error(
```
lucamancosu: we should not initialize or call a logger inside the domain. The logger is a specific of the implemententation i think. Maybe this logging should occur inside the event-collector module. 

PS: Probably we should avoid the initializzation of a logger instance inside the event-collector module(and other modules, we need a etl-func refactor for the other modules). We can:
1. just propagate the error to the entry point and log it 
2. OR
3. pass to the module an instance of the logger

apps/etl-func/src/domain/event.ts
```diff
@@ -1,3 +1,16 @@
...
 export interface EventProducer<T> {
   publish: (events: T[]) => Promise<void>;
 }
+
+export interface EventCollector<T> {
+  collect: (events: T[]) => Promise<void>;
+}
+
+export const eventsSummarySchema = z.object({
+  count: z.number().min(0),
```
lucamancosu: this should be min(1) ? an event sammury is created with at least 1 event count

apps/etl-func/local.settings.json.example
Garma00: We want to use a different cosmos here, we can use the same for test purpose so I left the same in the example file

apps/etl-func/local.settings.json.example
Garma00: ditto

Garma00: ditto

apps/etl-func/src/domain/event.ts
Garma00: [5fdc8de](https://github.com/pagopa/io-messages/pull/85/commits/5fdc8deb17b41029d44f23ed9cb2fc75e42a10c7)

apps/etl-func/src/adapters/cosmos/config.ts
Garma00: [b086645](https://github.com/pagopa/io-messages/pull/85/commits/b0866457e0c6af3c6f3fd1193b3bdbc81f4ffb22)

apps/etl-func/src/domain/use-cases/ingest-message.ts
Garma00: [dc87aa9](https://github.com/pagopa/io-messages/pull/85/commits/dc87aa9fc4e5c599fd6e96c3b9411aaf59fe7d40)

apps/etl-func/src/adapters/cosmos/__tests__/cosmos-collector.test.ts
```diff
@@ -0,0 +1,138 @@
...
+import { pino } from "pino";
+import { beforeEach, describe, expect, test, vi } from "vitest";
+
+import { CosmosWeeklyEventCollector } from "../event-collector.js";
+
+const patchMock = vi.fn();
+const readMock = vi.fn();
+const createMock = vi.fn();
+
+const logger = pino();
```
lucamancosu: You can apply something like 

const loggerErrorSpy = vi.spyOn(logger, "error").mockImplementation(() => {
  return;
});
to make logs "silent" during the tests

apps/etl-func/src/adapters/functions/__tests__/messages-ingestion.test.ts
```diff
@@ -5,17 +5,43 @@ import {
...
+    read: readMock,
+  }),
+  items: {
+    create: createMock,
+  },
+} as unknown as Container;
+
 const logger = pino();
+
+const cosmosCollectorMock = new CosmosWeeklyEventCollector(
```
lucamancosu: Should we check in each test if it is called ?

apps/etl-func/src/func.ts
```diff
@@ -59,9 +61,17 @@ const main = async (config: Config) => {
   );
   const messageAdapter = new MessageAdapter(blobMessageContentProvider, logger);
 
+  const ingestionSummaryContainer = new CosmosClient({
```
lucamancosu: Should we verfiy the cosmos connection inside the health route ?

apps/etl-func/local.settings.json.example
lucacavallaro, line 19: nitpick: i would rename `COSMOS_` in `COMMON_COSMOS_` to mark that `IOCOM_COSMOS` will be the default

apps/etl-func/src/adapters/cosmos/__tests__/cosmos-collector.test.ts
lucacavallaro: @Garma00 @lucamancosu the best way to make `pino` silent is 

```js
const logger = pino({ level: "silent" });
```

https://github.com/pinojs/pino/blob/main/docs/api.md#pinooptions-destination--logger

----

`spyOn` is useful if you want to test the logger/mock the implementation

apps/etl-func/src/adapters/cosmos/__tests__/cosmos-collector.test.ts
```diff
@@ -0,0 +1,138 @@
...
+  logger,
+);
+
+const eventsMock = [{ foo: "bar" }];
+
+beforeEach(vi.resetAllMocks);
+
+describe("CosmosWeeklyEventCollector.collect", () => {
+  test("when the cosmos read fails it should thwor an error", async () => {
+    readMock.mockReturnValueOnce(
```
lucacavallaro: Consider using [mockRejectedValueOnce](https://vitest.dev/api/mock#mockrejectedvalueonce) and its variant instead `mockReturnValue` or `mockImplementation`

apps/etl-func/src/adapters/functions/__tests__/messages-ingestion.test.ts
Garma00: In every test I call `cosmosCollectorMock.collect` inside the `test` case

apps/etl-func/src/adapters/cosmos/__tests__/cosmos-collector.test.ts
Garma00: Thanks, I made the logger silent as you suggested here [eba5465](https://github.com/pagopa/io-messages/pull/85/commits/eba5465217139707837340036c989d2982cee2e2)

apps/etl-func/src/func.ts
Garma00: You are right, take a look here [69d955b](https://github.com/pagopa/io-messages/pull/85/commits/69d955b6e57be5e6b59c359fdd856a4816c29a90)

apps/etl-func/local.settings.json.example
Garma00: [1f351ac](https://github.com/pagopa/io-messages/pull/85/commits/1f351ac5427632f95fb55298a7ca3ec76461c41d). [c09e1ba](https://github.com/pagopa/io-messages/pull/85/commits/c09e1ba05d22f84d85d5a09ba444029d23aa596a)

apps/etl-func/src/adapters/cosmos/__tests__/cosmos-collector.test.ts
Garma00: Thanks for the suggestion, take a look here [a3cb623](https://github.com/pagopa/io-messages/pull/85/commits/a3cb62366efc29589b16a25204d75b6d77be6566)

apps/etl-func/src/func.ts
lucamancosu: can we use this instead     await this.#client.getDatabaseAccount(); ?

apps/etl-func/src/adapters/cosmos/__tests__/cosmos-collector.test.ts
lucamancosu: const logger = pino({ level: "silent" })

Using this form i think we are not able to test the logger for checking if the logger is called when an error occurs for example. Make sense ?

I think is bettere to spy it. But is not a blocker

apps/etl-func/src/func.ts
Garma00: [1f32b6a](https://github.com/pagopa/io-messages/pull/85/commits/1f32b6a66d5a00ae4bf3962f89be39e2bf0c94a2)

infra/resources/_modules/web_apps/locals.tf
```diff
@@ -4,26 +4,29 @@ locals {
...
+      REDIS_PING_INTERVAL                            = 1000 * 60 * 9
+      REDIS_URL                                      = var.redis_cache.url
+      COMMON_COSMOS__accountEndpoint                 = var.cosmosdb_account_api.endpoint
+      COMMON_COSMOS_DBNAME                           = "db",
+      COMMON_COSMOS_MESSAGES_CONTAINER_NAME          = "messages-dataplan-ingestion-test"
+      COMMON_COSMOS_MESSAGE_STATUS_CONTAINER_NAME    = "message-status"
+      MESSAGE_ERROR_TABLE_STORAGE_NAME               = "MessagesDataplanIngestionErrors",
+      IOCOM_COSMOS__accountEndpoint                  = var.io_com_cosmos.endpoint
+      IOCOM_COSMOS_EVENTS_COLLECTOR_DBNAME           = "data-lake-cosmos-01"
+      IOCOM_COSMOS_INGESTION_SUMMARY_COLLECTION_NAME = "messages-summary"
```
lucamancosu, line 27: This collection already exists on Cosmos ? If not i propose to:

1. create it using terraform
2. rename it on something like "messages-ingestion-summary"

Garma00: This container already exist, it was created [here](https://github.com/pagopa/io-messages/pull/88/files#diff-31409a14222457666ea17358378aa3c124b61c9616fd67b3d4c30f2ada86834dR46)

lucamancosu: Perfect. Do you think a rename like messages-ingestion-summary (or statistics) can help us ? my "concern" is that messages-summary can be usefull as name for future scenarios. What do you think ?

Garma00: This container is inside a specific database called `data-lake`so I think that `ingestion` is superfluous, what's your take?

apps/etl-func/src/adapters/cosmos/__tests__/cosmos-collector.test.ts
Garma00: @lucamancosu  I agree, Take a look [eabc164](https://github.com/pagopa/io-messages/pull/85/commits/eabc164ee1ece86df3293bdfa2bca853d7b27206).

PS: I also left level `silent` so if we remove the spy in future we don't see logs in console during unit test.

lucamancosu (approved):

apps/etl-func/src/adapters/cosmos/event-collector.ts
```diff
@@ -0,0 +1,103 @@
...
+  EventsSummary,
+  eventsSummarySchema,
+} from "@/domain/event.js";
+import { Container, ItemResponse } from "@azure/cosmos";
+import { Logger } from "pino";
+
+/**
+ * Adapter for a cosmos weekly event collector.
+ * The goal is to store into a cosmos container the number of events sent into a
+ * container with year as partition key and week number as model id.
```
acuiuli:
```suggestion
 * The goal is to store the number of events sent to the event hub
 *  into a container with year as partition key and week number as model id.
```

apps/etl-func/src/domain/use-cases/ingest-message.ts
```diff
@@ -38,5 +41,11 @@ export class IngestMessageUseCase {
       ),
     );
     if (messagesEvent.length) await this.#eventProducer.publish(messagesEvent);
+
+    try {
+      this.#eventSummaryCollector.collect(messagesEvent);
+    } catch (error) {
+      return;
```
acuiuli: What happens if the collect fails? Did we analyze this case?

apps/etl-func/src/func.ts
```diff
@@ -85,9 +87,19 @@ const main = async (config: Config) => {
...
+  const ingestionSummaryContainer = ioComCosmosClient
+    .database(config.iocomCosmos.eventsCollectorDatabaseName)
+    .container(config.iocomCosmos.messageIngestionSummaryContainerName);
+
   const ingestMessageUseCase = new IngestMessageUseCase(
     messageAdapter,
     tokenizerClient,
+    new CosmosWeeklyEventCollector(ingestionSummaryContainer, logger),
     messageEventProducer,
   );
```
acuiuli:
```suggestion

  const eventCollector = new CosmosWeeklyEventCollector(ingestionSummaryContainer, logger);

  const ingestMessageUseCase = new IngestMessageUseCase(
    messageAdapter,
    tokenizerClient,
    eventCollector,
    messageEventProducer,
  );
```

Why not initializing it before as all the other parameters?

apps/etl-func/src/domain/use-cases/ingest-message.ts
Garma00: It is a fire and forget so we do not retry the ingestion. There is a log inside the collect method and I plan to convert it into an event after we merge #92.
I made a small refactor [51f634e](https://github.com/pagopa/io-messages/pull/85/commits/51f634e01eca92646ca18b4c64285d42fba14e09) as we discuss.

apps/etl-func/src/func.ts
Garma00: [871c047](https://github.com/pagopa/io-messages/pull/85/commits/871c047ad6c032b91ac30903231cfe2b7efbbab2)

acuiuli (approved):

PR #86, opened 2025-01-17
Iocom 2053 Ingestion error handling

lucamancosu:
We must store all the messages that returns an error during the ingestion process.

#### List of Changes

Augmented the ingestion logic for handle messages in errors.
Added an error table storage.

#### Motivation and Context

If a message cannot be ingest due to a specific error than we must be able to analyze it and take suitable actions.

I've tested it by throwing errors inside the ingest process chain. I've used the messagesMetadata and content available in the azure development environment.
I've created a table inside the azure development environment.

#### Types of changes

#### Checklist:

- [x] My change requires a change to the documentation.


apps/etl-func/docs/messages-ingestion/messages-ingestion-error-handling.puml
```diff
@@ -0,0 +1,30 @@
...
+    note left of etlFunc
+     The error can be throw while processing
+     while processing a batch or a single message metadata
+    end note
+    etlFunc -> ErrorQueueStorage: send message metadata
+    note left of etlFunc
+        Each item of the batch is sent individually to the queue
+    end note
+  else The function process the batch without throwing an error
+    etlFunc -> EventHub: Send messages to the eventhub
```
Garma00, line 24:
```suggestion
    etlFunc -> EventHub: Send messages to the eventhub
    deactivate etlFunc
```

apps/etl-func/src/adapters/config.ts
```diff
@@ -12,6 +12,10 @@ export const configSchema = z.object({
     messageStatusContainerName: z.string().min(1),
     messagesContainerName: z.string().min(1),
   }),
+  errorQueueStorage: z.object({
+    connectionString: z.string().url(),
```
Garma00: We are not using any connection string

apps/etl-func/src/adapters/functions/message-ingestion-error-queue.ts
```diff
@@ -0,0 +1,42 @@
...
+    //converting single object to an array to use the same ingestion logic
+    //it will always be an 1 item array
+    const queueItemArray = [...[], queueItem];
+    const parsedMessagesMetadata = queueItemArray.map(processMessageMetadata);
+    const messagesMetadata: MessageMetadata[] = parsedMessagesMetadata.filter(
+      (item): item is MessageMetadata => item !== undefined,
+    );
+
+    try {
+      await ingestUseCase.execute(messagesMetadata);
```
Garma00: I understand that you want to create an array in order to reuse the ingestUseCase. On the other hand, if we want to use this approach, I think it would be cleaner to parse the single item first and then create the array.
This way, you don't have to call any map. You can also remove the filter if you perform the check beforehand. I would write something similar.

```suggestion
const messagesIngestionErrorQueueHandler =
  (ingestUseCase: IngestMessageUseCase): StorageQueueHandler =>
  async (queueItem: unknown) => {
    const parsedMetadata = processMessageMetadata(queueItem);
     if (parsedMetadata === undefined) return;

    try {
      await ingestUseCase.execute([parsedMetadata]);
```

apps/etl-func/src/adapters/functions/messages-ingestion.ts
```diff
@@ -34,7 +33,13 @@ const messagesIngestionHandler =
     try {
       await ingestUseCase.execute(messagesMetadata);
     } catch (err) {
-      logger.error(`Error during the ingestion process`);
+      if (
+        context.retryContext?.retryCount === context.retryContext?.maxRetryCount
+      ) {
+        documents.forEach(
+          async (document) => await eventErrorRepository.push(document),
+        );
```
Garma00: Why are we pushing documents synchronously?
If we can't use a `batch` method I would do something similar

```suggestion
        Promise.all(
          documents.map((document) => eventErrorRepository.push(document)),
        );
```

infra/resources/_modules/storage_accounts/storage_account_api.tf
```diff
@@ -53,3 +53,14 @@ resource "azurerm_storage_table" "validationtokens" {
...
+
+resource "azurerm_storage_queue" "messages-dataplan-ingestion-errors" {
+  name                 = "messages-dataplan-ingestion-errors"
+  storage_account_name = module.storage_api.name
+}
+
+resource "azurerm_storage_queue" "messages-dataplan-ingestion-errors-poison" {
+  name                 = "messages-dataplan-ingestion-errors-poison"
+  storage_account_name = module.storage_api.name
+}
```
Garma00: Do we need a poison? I think that the queue we are creating already acts like a poison

Garma00 (changes requested):

infra/resources/_modules/storage_accounts/storage_account_api.tf
lucamancosu: the idea is:
1. CosmosDB trigger has a retry logic for batches. If the retries of a batch are not enough than the function push every single document(message metadata) to the queue. In this way we are able to send a single message metadata to the ingestion and isolate, if there is one, the message metadata that returns an error during the process.
2. Every run of a storageQueue trigger try to ingest a single message metadata. If it fails, after 5 retries, to send the message than the storageQueue trigger automatically put it inside the poison queue.
3. The poison is used  to understand how many messages are in error and as a storage where we can analyze the message problem.

So the queue is used to temporally store the single message metadata while the related poison queue is used to stored the message metadata that cannot be ingested.

apps/etl-func/docs/messages-ingestion/messages-ingestion-error-handling.puml
lucamancosu: 2a0ced619b90256f3803819eed888f5f8731b29e

apps/etl-func/src/adapters/config.ts
lucamancosu: 2a0ced619b90256f3803819eed888f5f8731b29e

apps/etl-func/src/adapters/functions/message-ingestion-error-queue.ts
lucamancosu: 2a0ced619b90256f3803819eed888f5f8731b29e

apps/etl-func/src/adapters/functions/messages-ingestion.ts
lucamancosu: 2a0ced619b90256f3803819eed888f5f8731b29e

apps/etl-func/docs/messages-ingestion/messages-ingestion-error-handling.puml
```diff
@@ -0,0 +1,31 @@
...
+'     A batch of is sent to the ETL function
+' end note
+ErrorQueueStorage -> etlFunc: Message metadata
+
+
+loop Process Mesagges Metadata
+  alt The function throw an error
+    note left of etlFunc
+     The error can be throw while processing
+     while processing a batch or a single message metadata
```
acuiuli:
```suggestion
     The error can be throw while processing
      a batch or a single message metadata
```

acuiuli (approved):

apps/etl-func/docs/messages-ingestion/messages-ingestion-error-handling.puml
lucamancosu: 69f0a3aa8131bcb22aded034c54fa6f081998901

acuiuli (approved):

apps/etl-func/src/adapters/config.ts
```diff
@@ -12,6 +12,10 @@ export const configSchema = z.object({
     messageStatusContainerName: z.string().min(1),
     messagesContainerName: z.string().min(1),
   }),
+  errorTableStorage: z.object({
+    connectionUri: z.string().url(),
+    tableName: z.string().min(1),
+  }),
```
Garma00: Just a consideration that we can consider in future as well, I would use a more specific name here at least for `tableName`, we are going to add another table for message-status ingestion errors and so we can get confused.
I would use a more specific name for `errorTableStorage` as well, something like `ingestionErrorsTableStorage`

apps/etl-func/src/adapters/functions/messages-ingestion.ts
```diff
@@ -1,42 +1,55 @@
...
+      const result = messageMetadataSchema.safeParse(document);
+      if (!result.success) {
+        invalidDocuments.push(document);
+        return;
+      }
+      if (!result.data.isPending) {
+        parsedMessagesMetadata.push(result.data);
+      }
+    });
+    let success = false;
```
Garma00: I would avoid this pattern in order to respect the principle of immutable data in functional programming. I am aware that this approach requires higher computational costs, but I would prefer something like
```suggestion
    const parsedMessagesMetadataOrZodError = documents.map((input) => {
    const parsedInput = messageMetadataSchema.safeParse(input);
    if (parsedInput.success) return parsedInput.data;
    else return parsedInput.error;
  })
  
  // filter the array in order to get only valid MessageMetadata with
  // isPending = false
  const nonPendingMessageMetadata = parsedMessagesMetadataOrZodError.filter(
    (input): input is MessageMetadata =>
      !(input instanceof ZodError) && !input.isPending,
  );

  // get all malformed documents so we can log them
  const malformedDocuments = parsedMessagesMetadataOrZodError.filter(
    (input): input is ZodError => input instanceof ZodError,
  );    
```
This way you can pass `nonPendingMessageMetadata` to the use case.
What's your take? @lucamancosu

apps/etl-func/src/adapters/table-storage/event-error-table-storage.ts
```diff
@@ -0,0 +1,45 @@
+import { EventErrorRepository } from "@/domain/event.js";
+import { TableClient } from "@azure/data-tables";
+import { v4 as uuidv4 } from "uuid";
+import { z } from "zod";
+
+const messageErrorTableEntitySchema = z.object({
+  errorType: z.string().min(1).optional(),
```
Garma00, line 7: Do we need this `errorType`? If so I would specify all possible cases with an `enum`

apps/etl-func/src/adapters/functions/__tests__/messages-ingestion.test.ts
```diff
@@ -54,6 +57,9 @@ const producer = new EventHubEventProducer(
   eventHubProducerClient,
   messageSchema,
 );
+
+const eventErrorRepository = new mocks.EventErrorRepository();
+
```
Garma00: I would update all the test cases in this file now that we have added a new component that impact the business logic.

- `shoud resolve if nothing throws` we should check that the `push` is not called
- `shoud call multiple times the business logic function if documents are more than one` we should check that the `push` is not called
- `shoud resolve if the documents array has malformed objects` we should check that the `publish` is called  exactly 1 time
- `should not call any function if the documents array is empty` we should check that the `publish` is not called
- `should throw an error if tokenize throws an error` we should check that the `publish` is called exactly 1 time
- `should resolve calling tokenize and getMessageByMetadata onfly for documents with isPending = false` we should check that the `publish` is not called

Garma00 (changes requested):

apps/etl-func/src/adapters/functions/messages-ingestion.ts
lucamancosu: Yes i like to use the same paradigm we have used in the ingestion process. @acuiuli and I were worried about the computational costs, and ingestion time, so we have did a try pushing the code like that.

I know about the functional requirement and there is no problem for me. @acuiuli What do you think ?

apps/etl-func/src/adapters/table-storage/event-error-table-storage.ts
lucamancosu: Right now we are using it for two macro types:
1. the message is malformed
2. the ingestion use case has failed for a batch of messages.
This two kind of error can help us debugging each message stored in the table storage. 
We are simply passing hardcoded string now. 

We can create enum error types for that but i think is better to not add it to the domain because they are related to te table storage implementation. We can than use this enum for add new possible cases in the future

apps/etl-func/src/adapters/functions/messages-ingestion.ts
acuiuli: We did it like this as Luca said to have it more optimized since the computational cost is already high. If we prefer to do it in a more functional way I can modify it.

acuiuli: modified [here](https://github.com/pagopa/io-messages/pull/86/commits/649dafa6aaff455943f9fa4e5a330859996ff197)

apps/etl-func/src/adapters/config.ts
lucamancosu: 7fceaa0699adea9228ee56e178ce0fc462495c37

apps/etl-func/src/adapters/functions/__tests__/messages-ingestion.test.ts
lucamancosu: 01a94e1c4205786e609900bae9d0905bc69f7b31

apps/etl-func/src/adapters/table-storage/event-error-table-storage.ts
lucamancosu: b9fbc862f0c692cf3f1053eabb828bd4687b0a6f

apps/etl-func/src/adapters/functions/messages-ingestion.ts
```diff
@@ -1,42 +1,64 @@
...
+      (document): document is MessageMetadata =>
+        !(document instanceof ZodError) && !document.isPending,
     );
 
+    // get all malformed documents so we can send them to the error repository
+    const malformedDocuments = parsedMessagesMetadataOrZodError.filter(
+      (input): input is ZodError => input instanceof ZodError,
+    );
+
+    let success = false;
```
Garma00: I think we should remove this flag here and from the if statement

Garma00 (changes requested):

apps/etl-func/src/adapters/functions/messages-ingestion.ts
lucamancosu: 53c55eea9b79646a70fd497ac0f83d709be97c04

I've changed the malformedDocuments push too. I've introduced a promise all with a .map instead of a foreach

Garma00 (approved):

acuiuli (approved):

PR #92, opened 2025-02-05
Error management for message status ingestion

acuiuli:
#### List of Changes

- Added error management logic with a table storage to store all the message status that returns an error
- Created a new table storage
- Evolved tests

#### Motivation and Context

We need to store all the message status that returns an error during the ingestion process in order to be able to analyze it and recover.

#### Types of changes

#### Checklist:


apps/etl-func/src/adapters/functions/message-status-ingestion.ts
```diff
@@ -1,20 +1,73 @@
...
+    const parsedMessageStatusesOrZodError = documents.map((input) => {
+      const parsedInput = messageStatusSchema.safeParse(input);
+      if (parsedInput.success) return parsedInput.data;
+      else return parsedInput.error;
     });
-    const messageStatusBatch = parsedMessageStatusBatch.filter(
-      (status): status is MessageStatus => status !== undefined,
+
+    // filter the array in order to get only valid MessageMetadata with
+    // isPending = false
```
Garma00:
```suggestion
    // filter the array in order to get only valid MessageStatus
```

Garma00 (approved):

PR #97, opened 2025-02-11
chore: moving cosmosdb trigger setting to the env vars

lucamancosu:
Moving hard coded cosmosdb trigger settings to the env vars.

#### List of Changes

Added a few env vars for the cosmosdb trigger

#### Motivation and Context

Make easier to change trigger settings while testing on the target environment

Running the etl function locally and completing an ingestion and a failed ingestion flow.
#### Types of changes

#### Checklist:


apps/etl-func/local.settings.json.example
```diff
@@ -22,8 +22,13 @@
     "COMMON_COSMOS_MESSAGE_STATUS_CONTAINER_NAME": "message-status",
     "MESSAGE_ERROR_TABLE_STORAGE_NAME": "MessagesDataplanIngestionErrors",
     "ACCOUNT_STORAGE__tableServiceUri": "https://testgiovannist.table.core.windows.net/",
-    "APPINSIGHTS_CONNECTION_STRING": "aConnectionString",
-    "APPINSIGHTS_SAMPLING_PERCENTAGE": 100,
+    APPINSIGHTS_CONNECTION_STRING: "aConnectionString",
+    APPINSIGHTS_SAMPLING_PERCENTAGE: 100
```
lucacavallaro: have you removed double quotes?

Garma00:
```suggestion
    "APPINSIGHTS_CONNECTION_STRING": "aConnectionString",
    "APPINSIGHTS_SAMPLING_PERCENTAGE": 100
```

apps/etl-func/src/adapters/env.ts
```diff
@@ -18,6 +18,11 @@ export const envSchema = z.object({
   MESSAGE_EVENTHUB_NAME: z.string().min(1),
   MESSAGE_STATUS_ERROR_TABLE_STORAGE_NAME: z.string().min(1),
   MESSAGE_STATUS_EVENTHUB_NAME: z.string().min(1),
+  MESSAGES_INGESTION_TRIGGER_LEASE_CONTAINER: z.string().min(1),
+  MESSAGES_INGESTION_TRIGGER_MAX_INVOCATION_ITEMS: z.coerce.number(),
+  MESSAGES_INGESTION_TRIGGER_MAX_RETRY_COUNT: z.coerce.number(),
+  MESSAGES_INGESTION_TRIGGER_MAXIMUM_INTERVAL: z.coerce.number(),
+  MESSAGES_INGESTION_TRIGGER_MINIMUM_INTERVAL: z.coerce.number(),
```
lucacavallaro: Do we really need those variables? I don't understand the purpose of configuring these values via env var

also the schemas are too generic, is `5.23948923` valid for `MAX_RETRY_COUNT`? why aren't there any default values?

lucacavallaro (changes requested):

lucacavallaro:
@lucamancosu could you change the PR title so that it’s clearer and properly formatted?

apps/etl-func/local.settings.json.example
```diff
@@ -22,8 +22,13 @@
...
     "MESSAGE_STATUS_ERROR_TABLE_STORAGE_NAME": "MessageStatusDataplanIngestionErrors",
+    "MESSAGES_INGESTION_TRIGGER_LEASE_CONTAINER": "messages-dataplan-ingestion-test-lease",
+    "MESSAGES_INGESTION_TRIGGER_MAX_INVOCATION_ITEMS": 50,
+    "MESSAGES_INGESTION_TRIGGER_MINIMUM_INTERVAL": 3,
+    "MESSAGES_INGESTION_TRIGGER_MAXIMUM_INTERVAL": 60,
+    "MESSAGES_INGESTION_TRIGGER_MAX_RETRY_COUNT": 5
```
Garma00: Just a consideration, `TRIGGER` can be removed cause it doesn't add anyi information.
I would give information about time unit such as `MESSAGES_INGESTION_MINIMUM_INSTERVAL_AS_SECONDS`, what's your take @lucamancosu ?

Garma00 (changes requested):

apps/etl-func/src/adapters/env.ts
lucamancosu: The main idea is to have the variables available on azure portal while testing the function if we need to change them. 

Yes you are right, is better to have a more detailed schema.

lucacavallaro: Do we need to change them? When and why? I don't think we will change them frequently in production.
I fear that having too many variables will only make the project harder to configure.

lucamancosu: The idea is to use a 1 minute time range for test the retry logic on the target environment. 
After the test is finished we must update it to 30 minutes.

If we want to avoid these new variables we can do the update simply changing the hardcoded values and doing a new function deploy. In this last scenario we can archive this PR

apps/etl-func/local.settings.json.example
lucamancosu: f3e793d0b4196e6f94e82c87db64f0fa92b10110

lucamancosu: f3e793d0b4196e6f94e82c87db64f0fa92b10110

apps/etl-func/local.settings.json.example
lucamancosu: just a litle rename here 185e7f0b92de02d95259d028010e64ec1eab3e7a

Garma00 (approved):

Garma00 (approved):

lucacavallaro (approved):

PR #107, opened 2025-02-14
refactor(infra): add new storage account and table

Garma00:
#### List of Changes

* Add a new storage account for the domain (only tables are enabled for now);
* Add a new table to collect messages ingestion errors
* Update etl role in order to use the new table instead of the old one
* Add autoscale_settings for the `messages-summary` datalake container

#### Motivation and Context

We need a table where to collec messages ingestion errors, we want to put it in a new 
storage account dedicated for the domain.
We already have a table in a common storage account that can be removed once we create this.

#### Types of changes

#### Checklist:


infra/resources/_modules/com_storage_account/storage_account_io_com.tf
```diff
@@ -0,0 +1,22 @@
+module "dx-azure-storage-account" {
+  source  = "pagopa/dx-azure-storage-account/azurerm"
+  version = "0.0.9"
+
+  subnet_pep_id       = var.subnet_pep_id
+  tags                = var.tags
+  tier                = "s"
+  environment         = merge(var.environment, { app_name = "messages" })
```
lucacavallaro: I would remove the `app_name` to get `iopitncomst` as resource name (if `app_name` can't be empty, remove the `domain` and set `com` as `app_name`)

infra/resources/_modules/com_storage_account/storage_account_io_com.tf
```diff
@@ -0,0 +1,22 @@
+module "dx-azure-storage-account" {
```
lucacavallaro:
```suggestion
module "com_st" {
```

infra/resources/_modules/com_storage_account/main.tf
lucacavallaro, line 1: why did you create a new module for this storage account (`com_storage_account`) ? we already have `storage_accounts`

infra/resources/prod/main.tf
```diff
@@ -232,3 +232,17 @@ module "cosmos" {
     instance_number = "01"
   }
 }
+
+module "io_com_storage_account" {
+  source              = "../_modules/com_storage_account/"
+  resource_group_name = data.azurerm_resource_group.itn_messages.name
```
lucacavallaro: we should use `itn_com` resource group (`io-p-itn-com-rg-01`) for new resources

infra/resources/prod/main.tf
```diff
@@ -232,3 +232,17 @@ module "cosmos" {
...
+  environment = {
+    prefix          = local.prefix
+    env_short       = local.env_short
+    domain          = local.domain
+    location        = local.location
+    instance_number = "01"
```
lucacavallaro: `instance_number` is a detail that can be defined inside the module

infra/resources/_modules/com_storage_account/variables.tf
lucacavallaro, line 1: if you reuse `storage_accounts`, you won't need to replicate the same variables everywhere!

lucacavallaro (changes requested):

infra/resources/_modules/com_storage_account/main.tf
Garma00: @lucacavallaro  As I mentioned in the description I tried to define this new storage account inside the `storage_accounts` folder but it is a bit messy using the new dx module.
I need to add some more variables that are not needed for the storage already defined in that module so I thought it was a bit more cleaner to define another module.

infra/resources/_modules/com_storage_account/variables.tf
Garma00: @lucacavallaro I got some problems trying to reuse that module because I want to use the new dx module which requires some variables that are not needed inside `storage_accounts`.
For example I need to add a new pep_id which is not requested by the `storage_api_weu`, so I would need to add pep_id variable and pass it [here](https://github.com/pagopa/io-messages/blob/main/infra/resources/prod/westeurope.tf#L23) even thought it will not use it.

My proposal is to use this new module `com_storage_account` and unify those two modules once we migrate the old storage account to the new dx module.

infra/resources/prod/main.tf
Garma00: [8640ea0](https://github.com/pagopa/io-messages/pull/107/commits/8640ea092a73f233ba5b83b69acbb1ce730c596e)

infra/resources/_modules/com_storage_account/storage_account_io_com.tf
Garma00: [3d45663](https://github.com/pagopa/io-messages/pull/107/commits/3d45663e42a77af74c6591ed4cf6f59b2e599250)

infra/resources/_modules/com_storage_account/storage_account_io_com.tf
Garma00: [8ecfb59](https://github.com/pagopa/io-messages/pull/107/commits/8ecfb593036761f7de00d340f4bfbc37c29ee9bb)

infra/resources/prod/main.tf
Garma00: [3d45663](https://github.com/pagopa/io-messages/pull/107/commits/3d45663e42a77af74c6591ed4cf6f59b2e599250)

infra/resources/_modules/web_apps/locals.tf
```diff
@@ -1,9 +1,9 @@
 locals {
   location_short = var.environment.location == "italynorth" ? "itn" : var.environment.location == "westeurope" ? "weu" : var.environment.location == "germanywestcentral" ? "gwc" : "neu"
-  project        = "${var.environment.prefix}-${var.environment.env_short}-${local.location_short}"
   etl_func = {
     app_settings = {
       NODE_ENV                                       = "production",
+      "AzureWebJobs.IngestMessages.Disabled"         = true,
```
lucamancosu, line 6: Should we remove the comments on the message status trigger and add here the same Disbled property ?

Garma00: I guess but I would do this into another pr related to the message-status ingestion

lucamancosu (approved):

lucacavallaro (approved):

PR #121, opened 2025-02-19
Set keyless CosmosDB connection in `sending-func`

lucacavallaro:
Deprecate `Key` as connection method


Garma00 (commented):
@lucacavallaro I think that we can add a call to those 2 cosmos accounts in the healthcheck function in order to stop the deploy in case of problems, what's your take?

For example we are using [this](https://github.com/pagopa/io-messages/blob/main/apps/etl-func/src/func.ts#L145) in the etl functio

lucacavallaro:
> @lucacavallaro I think that we can add a call to those 2 cosmos accounts in the healthcheck function in order to stop the deploy in case of problems, what's your take?
> 
> For example we are using [this](https://github.com/pagopa/io-messages/blob/main/apps/etl-func/src/func.ts#L145) in the etl functio

Done https://github.com/pagopa/io-messages/pull/121/commits/b63f2c6e25219301cf087284e7e145bf0bad8da0

Garma00 (approved):

PR #130, opened 2025-02-20
Change summary handling

Garma00:
At the moment message ingestion's summaries are collected in a cosmos container. The same record is updated for batches that are processed within a week.
This handling cause errors due to race condition when there are more than 1 function instance.

To avoid this problem a new record for a batch will be created.


Garma00:
You can find on #85 the original implementation.
The original `summary` shape was:
```json
{
  "year": "2025"
  "id": "2025-W01"
  "count": 100
}
```

Now will look like this:
```json
{
  "year": "2025"
  "id": "01JMHVSXQ9K25K0WB3HH0DTDAG"
  "count": 100
}
```

Where the `id` is the id of the first message of the batch.

apps/etl-func/src/adapters/cosmos/event-collector.ts
```diff
@@ -1,63 +1,56 @@
...
+   * Accept a batch of events and return the summary.
+   * */
+  private createSummary(events: [T, ...T[]]): EventsSummary {
     return {
       count: events.length,
-      id: this.getCurrentModelId(),
+      id: this.getBatchId(events),
       year: this.getCurrentPartitionKey(),
     };
   }
```
lucacavallaro: why is this a method? we are creating a unnecessary coupling here

apps/etl-func/src/adapters/cosmos/event-collector.ts
```diff
@@ -1,63 +1,56 @@
...
+  /**
+   * Returns the id of the first event in the batch.
+   *
+   * @param events Accept a non empty array
+   *
+   * @returns The id of the first event
+   * */
+  private getBatchId(events: [T, ...T[]]): string {
+    return events[0].id;
   }
```
lucacavallaro: I would do this check at runtime (maybe using the `get` method of `Array`)

apps/etl-func/src/adapters/cosmos/event-collector.ts
```diff
@@ -1,63 +1,56 @@
...
-    return `${today.getFullYear()}-W${weekNumber.toString().padStart(2, "0")}`;
+  /**
+   * Returns the id of the first event in the batch.
+   *
+   * @param events Accept a non empty array
+   *
+   * @returns The id of the first event
+   * */
+  private getBatchId(events: [T, ...T[]]): string {
+    return events[0].id;
```
lucamancosu: There is a possibility that this can cause collisions ? If its possible we should use a new uuidv4 or a cuuid. In this way there is no need to pass the entire events array to the collect method but the events count only

Garma00: In our case we should never get the collision because every message is processed only once cc @acuiuli @lucacavallaro

apps/etl-func/src/adapters/cosmos/event-collector.ts
Garma00: [5a6d7ad](https://github.com/pagopa/io-messages/pull/130/commits/5a6d7ad7fb8475827c91e99444f8ddd2ccdfc48b)

apps/etl-func/src/adapters/cosmos/event-collector.ts
Garma00: I added inside the `collect` a runtime check in order to skip empty batches, we don't want to throw here because we would reprocess the entire batch [c230c9d](https://github.com/pagopa/io-messages/pull/130/commits/c230c9d392e911286bd57d5020b2ff2b2bfbcef4)

lucamancosu (approved):

apps/etl-func/src/adapters/cosmos/event-collector.ts
Garma00: We are using uuid now as you suggested [14e7f9f](https://github.com/pagopa/io-messages/pull/130/commits/14e7f9fab6ba0f3dcf4ea8812599fbdbd8f90ed5)

apps/etl-func/src/adapters/cosmos/event-collector.ts
```diff
@@ -1,63 +1,29 @@
-import {
-  EventCollector,
-  EventsSummary,
-  eventsSummarySchema,
-} from "@/domain/event.js";
+import { EventCollector, EventsSummary } from "@/domain/event.js";
 import { TelemetryEventName, TelemetryService } from "@/domain/telemetry.js";
 import { Container, ItemResponse } from "@azure/cosmos";
+import { v4 } from "uuid";
```
lucacavallaro: use `ulid` instead of `uuid` - it is sortable and it is the default choice for our IDs on IO

lucacavallaro: example in js

```js
import { ulid } from "ulid";

const newId = ulid();

const idSchema = z.string().ulid();
```

lucacavallaro (changes requested):

apps/etl-func/src/adapters/cosmos/event-collector.ts
Garma00: Ok! Take a look here [bb5417f](https://github.com/pagopa/io-messages/pull/130/commits/bb5417f45db223837c26504402cf7137b554c58a)

yarn.lock
```diff
@@ -6367,6 +6367,7 @@ __metadata:
     prettier: "npm:3.2.5"
     redis: "npm:^4.7.0"
     typescript: "npm:5.4.3"
+    ulid: "npm:^2.3.0"
     uuid: "npm:^11.0.5"
```
lucacavallaro, line 6371: is `uuid` still needed?

lucacavallaro (changes requested):

yarn.lock
Garma00: We are using it in table storage adapter

lucacavallaro (approved):

PR #139, opened 2025-02-21
Fix ingestion bugs

acuiuli:
- Fixed malformed document bug that wrote the ZodError instead of the malformed document in the table storage
- Modified table storage account uri with the correct one


apps/etl-func/src/adapters/functions/messages-ingestion.ts
```diff
@@ -16,23 +12,20 @@ const messagesIngestionHandler =
...
-    const nonPendingMessageMetadata = parsedMessagesMetadataOrZodError.filter(
-      (document): document is MessageMetadata =>
-        !(document instanceof ZodError) && !document.isPending,
-    );
+    const nonPendingMessageMetadata = documents
+      .filter((document) => {
+        const parsedInput = messageMetadataSchema.safeParse(document);
+        return parsedInput.success && !parsedInput.data.isPending;
+      })
+      .map((document) => messageMetadataSchema.parse(document));
```
Garma00, line 22: We can't directly filter here because we could have some elements in input without some `default` properties defined in the zod schema.
This will cause a `success` when calling the `safeParse` method but the element returnend in the array will be `document` and the default properties would miss.
That's why we had a previous `map` method.

apps/etl-func/src/adapters/functions/messages-ingestion.ts
```diff
@@ -16,23 +12,20 @@ const messagesIngestionHandler =
...
 
     // get all malformed documents so we can send them to the error repository
-    const malformedDocuments = parsedMessagesMetadataOrZodError.filter(
-      (input): input is ZodError => input instanceof ZodError,
-    );
+    const malformedDocuments = documents.filter((document) => {
+      const parsedInput = messageMetadataSchema.safeParse(document);
+      return !parsedInput.success;
+    });
```
Garma00, line 28: If you do a map before the filter as I wrote in the comment before, you can directly filter with a guard instead of calling the safeParse again.

apps/etl-func/src/adapters/functions/messages-ingestion.ts
```diff
@@ -58,17 +51,23 @@ const messagesIngestionHandler =
...
     if (malformedDocuments.length > 0) {
-      await Promise.all(
-        malformedDocuments.map((malformedDocument) =>
-          eventErrorRepository.push(
-            malformedDocument,
-            EventErrorTypesEnum.enum.MALFORMED_EVENT,
+      try {
+        telemetryService.trackEvent(TelemetryEventName.MALFORMED_MESSAGES, {
+          invocationId: context.invocationId,
+        });
```
Garma00: What about adding a `malformedCount` property here?

Garma00 (changes requested):

apps/etl-func/src/adapters/functions/messages-ingestion.ts
acuiuli: That's why we added the map step after the filter with the extra .parse, in order to have the default properties defined.

apps/etl-func/src/adapters/functions/messages-ingestion.ts
acuiuli: ditto

apps/etl-func/src/adapters/functions/messages-ingestion.ts
acuiuli: solved [here](https://github.com/pagopa/io-messages/pull/139#:~:text=to%20MALFORMED_MESSAGES%20event-,523411e,-acuiuli%20requested%20a)

Garma00 (approved):

PR #145, opened 2025-02-25
Reduce `messages-summary` CosmosDB Container RU

acuiuli:
Reduce the RU from 800 to 400


Garma00 (approved):

infra/resources/prod/web_apps.tf
```diff
@@ -31,7 +31,7 @@ module "web_apps" {
   nat_gateway_id = data.azurerm_nat_gateway.itn_ng.id
 
   app_settings = {
-    message_error_table_storage_uri : "https://${module.storage_api_weu.com_st_name}.table.core.windows.net",
```
lucacavallaro: Can you make this change in another PR?

lucacavallaro (changes requested):

infra/resources/prod/web_apps.tf
lucacavallaro: (why is this variable hardcoded? can't be obtained from `st_com` resource?)

lucacavallaro (approved):

PR #155, opened 2025-02-27
Add autoscaler to `etl-func`

lucamancosu:
Closes IOCOM-2196


infra/resources/_modules/web_apps/etl_func.tf
```diff
@@ -32,6 +32,33 @@ module "etl_func" {
...
 }
 
+module "etl_func_autoscaler" {
+  source  = "pagopa/dx-azure-app-service-plan-autoscaler/azurerm"
+  version = "~>0"
+
+  app_service_plan_id = module.etl_func.function_app.plan.id
+  location            = var.environment.location
+
+  autoscale_name      = "${var.environment.prefix}-${var.environment.env_short}-${local.location_short}-${var.environment.domain}-etl-func"
```
lucacavallaro: why are you specifying the `autoscale_name`? it can be omitted and it will be inferred by the `function_app` name

infra/resources/_modules/web_apps/etl_func.tf
```diff
@@ -32,6 +32,33 @@ module "etl_func" {
...
+    }
+  }
+
+  scheduler = {
+    maximum = 4
+    normal_load = {
+      default = 3
+      minimum = 1
+    }
+  }
```
lucacavallaro: can't we use the default scheduler? can you add a comment that describe why we decided to put this values?

lucacavallaro (changes requested):

infra/resources/_modules/web_apps/etl_func.tf
lucamancosu: We have successfully tested the function with 4 instances. The main bottleneck is PDV and to avoid rates limit problems i prefer to use a maximum of 4 instances and the minimum required by the function service plan for the normal load. 
If you agree with this i'll put a comment.

infra/resources/_modules/web_apps/etl_func.tf
lucamancosu: You're right. I've noticed now that is not mandatory 
86e3676987bc812859aab69cc264a02a50bf596f

infra/resources/_modules/web_apps/etl_func.tf
lucacavallaro: Agree. But you have also to document the choice to NOT configure a `low_load` profile.

lucamancosu: low profile requires a start and end time. In this moment we don't have statistics about the create or update message. So my proposal is to setup the low and max profile when we have sufficient data.

What do you think ?

lucacavallaro: Actually, we have this data; you can look for `SubmitMessage` or `CreateMessage` functions' metrics. The default schedule was set based on actual load data.

lucamancosu: Right now we want to take a minimum and a maximum of 4 instances until we align the ingestion with the nowadays' messages.

After we did it, we can insert the configurations for the low and max load.

dabe9b759e4e3358d1eead7b5c17d9bca2ae64fd

Garma00 (approved):

infra/resources/_modules/web_apps/etl_func.tf
lucacavallaro: Can you add the comment in the tf file?

lucacavallaro (approved):

infra/resources/_modules/web_apps/etl_func.tf
lucamancosu: done
def6e01fd59db4db107f00beb30dab2f8e0e0e7c

PR #156, opened 2025-02-27
Change messages container and ingestion lease container

lucamancosu:
Needed before start the production ingestion


infra/resources/_modules/cosmos/cosmos_etl_messages.tf
```diff
@@ -1,7 +1,7 @@
 module "comsosdb_sql_container_messages_ingestion_test" {
   source = "github.com/pagopa/terraform-azurerm-v4//cosmosdb_sql_container?ref=v1.2.1"
 
-  name                = "messages-dataplan-ingestion-test"
+  name                = "messages"
   resource_group_name = var.cosmosdb_account.resource_group_name
```
Garma00: I think this can be removed, we already have the messages container defined [here](https://github.com/pagopa/io-infra/blob/main/src/common/_modules/cosmos_api/cosmos_containers.tf)

Garma00 (changes requested):

infra/resources/_modules/cosmos/cosmos_etl_messages.tf
lucamancosu: ea9567bebdb5b1fc0c2fbc043c9e02c147eb8009

Garma00 (approved):

Garma00 (approved):

PR #158, opened 2025-02-27
Add terraform static analysis into github workflows

lucamancosu:


lucamancosu:
Must be merged only when https://github.com/pagopa/io-messages/pull/155 is completed

lucamancosu:
The action keeps failing because is creating/updating the files.
We've noticed that when running the pre-commit command on different machines, it always updates a few README files, modifying the azurerm version information inside those files. We tested it ensuring that we were using the same versions of Terraform, pre-commit, azurerm, etc.

My conclusion is that the pre-commit command updates the README files differently depending on the machine it is running on.

Do you have any suggestion @lucacavallaro  ?

Garma00:
No longer needed,

PR #161, opened 2025-02-28
Set default and minimum instances to 1 for etl autoscaler

Garma00:


Garma00:
The ingestion will start next week so we can set minimum instances to 1 till the start day.

acuiuli (approved):

PR #171, opened 2025-03-06
Update time_window on ingestion alerts

acuiuli:
Changed the ingestion alerts time windows to match the frequency in order to avoid multiple alerts for a single error

### DIFF
--- a/.changeset/green-radios-promise.md
+++ b/.changeset/green-radios-promise.md
@@ -1,5 +1,0 @@
----
-"etl-func": patch
----
-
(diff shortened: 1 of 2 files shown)


.changeset/green-radios-promise.md
lucacavallaro, line 1: changesets should be used only for changes in application code, our infrastructure is not versioned

lucacavallaro (changes requested):

lucacavallaro (approved):

Garma00 (approved):

PR #178, opened 2025-03-12
Update dx modules

lucacavallaro:
* Update dx modules to their latest version
* Change dx modules source to sue the new `pagopa-dx` org


christian-calabrese (approved):
LGTM

PR #181, opened 2025-03-13
Add ops-func

Garma00:
- Create a new function-app `ops-func`
- Create a new function `SplitDeleteMessageChunks`
- Create a new function `DeleteMessage`

Resolves IOCOM-2225


.yarnrc.yml
```diff
@@ -1,4 +1,5 @@
 preferReuse: true
+checksumBehavior: update
```
lucacavallaro: why do you set this option here?

apps/util-func/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "util-func",
```
lucacavallaro: I think `util-func` is too vague, I would rename it to `ops-func` to make more understandable the purpose of this service (exposing `operations` api).

apps/util-func/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "main": "dist/func.js",
+  "type": "module",
+  "files": [
+    "host.json"
+  ],
```
lucacavallaro: this field is no longer needed
```suggestion
```

apps/util-func/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "packageManager": "yarn@4.7.0",
```
lucacavallaro: why does this version is different that the one in the root `package.json`?

apps/util-func/src/adapters/blob-storage/message-content-deleter.ts
```diff
@@ -0,0 +1,41 @@
...
+      } else {
+        this.logger.error(
+          `Failed to delete message content for message ${messageId} | Error code: ${response.errorCode}`,
+        );
+      }
+    } catch (error) {
+      this.logger.error(
+        `Failed to delete message content for message ${messageId} | Error: ${error}`,
+      );
+    }
```
lucacavallaro: I would move the logger out of this class, IMHO the logging should be a consumer responsability

apps/util-func/src/adapters/config.ts
```diff
@@ -0,0 +1,8 @@
+import * as z from "zod";
+
+export const envSchema = z.object({
+  COM_STORAGE_ACCOUNT__serviceUri: z.string().url(),
```
lucacavallaro: I think our common resources (`COM` without a specific `app_name`) can be referenced without the prefix
```suggestion
  STORAGE_ACCOUNT__serviceUri: z.string().url(),
```

apps/util-func/src/domain/message-metadata.ts
```diff
@@ -0,0 +1,3 @@
+export interface MessageMetadataDeleter {
+  deleteMessageMetadata: (partitionKey: string, id: string) => Promise<void>;
+}
```
lucacavallaro: I think that these deleter should not be placed here - the domain known only about messages and should not act on their part independently

lucacavallaro (changes requested):

.yarnrc.yml
Garma00: Thanks for the catch, I had an error running the install locally but everything is fine now [994e333](https://github.com/pagopa/io-messages/pull/181/commits/994e333f330b7c443a0892ccdf3d8875a5e2405b)

apps/util-func/package.json
Garma00: I renament it into `ops-ca` [f7a5199](https://github.com/pagopa/io-messages/pull/181/commits/f7a519928f91dc9dd6b6a4fa84a362c23f692a39)

apps/util-func/package.json
Garma00: [5b4e53a](https://github.com/pagopa/io-messages/pull/181/commits/5b4e53a5880205d89002747ac85d2cfefe8909ed)

apps/util-func/package.json
Garma00: [8af941c](https://github.com/pagopa/io-messages/pull/181/commits/8af941c8d37643395b3169843b09c17a73338ddf)

apps/util-func/src/adapters/config.ts
Garma00: [890a673](https://github.com/pagopa/io-messages/pull/181/commits/890a673982f24123ec68a246975d2593bfb464fa)

apps/util-func/src/domain/message-metadata.ts
Garma00: [70778d8](https://github.com/pagopa/io-messages/pull/181/commits/70778d8bfe656be318d01855b0a5f16bd5e0b189)

apps/util-func/src/adapters/blob-storage/message-content-deleter.ts
Garma00: [2f9b688](https://github.com/pagopa/io-messages/pull/181/commits/2f9b688a12eea3a5c156c01389bd9204f406708f)

apps/util-func/package.json
lucacavallaro: I've thought about it, and I think you can remove this field entirely as it's only needed in the root workspace

Garma00: [7f66baf](https://github.com/pagopa/io-messages/pull/181/commits/7f66baf057a5e9d278f67a0abae6a741b00e0067)

apps/ops-ca/src/adapters/functions/delete-messages.ts
```diff
@@ -0,0 +1,21 @@
+import { DeleteMessageUseCase } from "@/domain/use-cases/delete-message.js";
+import { StorageBlobHandler } from "@azure/functions";
+
+export const deleteMessages =
+  (deleteMessageUseCase: DeleteMessageUseCase): StorageBlobHandler =>
+  async (blob, context) => {
+    if (!Buffer.isBuffer(blob)) {
+      context.error("Invalid input input blob file");
```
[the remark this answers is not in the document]
Garma00: [9dcf5d7](https://github.com/pagopa/io-messages/pull/181/commits/9dcf5d7bb1b035705171baec3cb3886ba96de58a)

infra/resources/_modules/web_apps/ops_ca.tf
```diff
@@ -0,0 +1,70 @@
...
+    COSMOS_URI                  = var.cosmosdb_account_api.endpoint,
+    COSMOS_DATABASE_NAME        = var.cosmosdb_account_api.name,
+    COMMON_STORAGE_ACCOUNT_URL  = var.message_content_storage.endpoint,
+    STORAGE_ACCOUNT__serviceUri = var.app_settings.message_error_table_storage_uri
+  }
+}
+
+# TODO: this function should be updated to use container app when the terraform
+# provider supports it
+module "ops_ca" {
```
lucacavallaro:
```suggestion
module "ops_func" {
```

Garma00: [4800033](https://github.com/pagopa/io-messages/pull/181/commits/48000331d1621f092f9d85c50ee9af3e3707c913)

apps/ops-func/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "ops-ca",
```
lucacavallaro:
```suggestion
  "name": "ops-func",
```

apps/ops-func/src/adapters/cosmos/message-metadata-deleter.ts
```diff
@@ -0,0 +1,50 @@
...
+  /**
+   * Deletes a message metadata from the container if it exists.
+   *
+   * @param fiscalCode - The partition key of the item.
+   * @param messageId - The id of the item.
+   * @returns A promise that resolves when the deletion is complete.
+   */
+  async deleteMessageMetadata(
+    fiscalCode: string,
+    messageId: string,
```
lucacavallaro: Can't we use branded types? we should have a `FiscalCode` type, maybe we need to add the `MessageId` type as well

apps/ops-func/src/adapters/functions/split-delete-messages.ts
```diff
@@ -0,0 +1,29 @@
...
+        context.error(
+          `Invalid pair fiscalCode: ${fiscalCode}, messageId: ${messageId}`,
+        );
+        return;
+      }
+
+      return queueClient.sendMessage(
+        Buffer.from(JSON.stringify({ fiscalCode, messageId })).toString(
+          "base64",
+        ),
```
lucacavallaro: this can be simplified using `queue output` (removing the need to use a custom queue logic)

apps/ops-func/src/adapters/functions/split-delete-messages.ts
```diff
@@ -0,0 +1,29 @@
...
+
+      if (!fiscalCode || !messageId) {
+        context.error(
+          `Invalid pair fiscalCode: ${fiscalCode}, messageId: ${messageId}`,
+        );
+        return;
+      }
+
+      return queueClient.sendMessage(
+        Buffer.from(JSON.stringify({ fiscalCode, messageId })).toString(
```
lucacavallaro: it would be better to have a separate function that serialize the message

apps/ops-func/src/adapters/functions/split-delete-messages.ts
```diff
@@ -0,0 +1,29 @@
...
+    if (!Buffer.isBuffer(blob)) {
+      context.error("Invalid input blob file");
+      return;
+    }
+    const lines = blob.toString("utf-8").trim().split("\n");
+
+    lines.forEach((line) => {
+      const [fiscalCode, messageId] = line.split(",");
+
+      if (!fiscalCode || !messageId) {
```
lucacavallaro, line 16: you should also parse these values to make sure that they follow the correct format

lucacavallaro (changes requested):

apps/ops-func/src/adapters/cosmos/message-metadata-deleter.ts
Garma00: I was thinking that a potential use case for this function could be handling messages with malformed messageId or FiscalCode, adding this strict check at this stage might prevent us from deleting it later. What's your take?

lucacavallaro: good point 👍 please explain this rationale in a comment

apps/ops-func/src/adapters/functions/split-delete-messages.ts
Garma00: As we discussed [here](https://github.com/pagopa/io-messages/pull/181#discussion_r2042247750) I think it is better to not check that the format is correct

apps/ops-func/src/adapters/cosmos/message-metadata-deleter.ts
Garma00: [5191884](https://github.com/pagopa/io-messages/pull/181/commits/519188479ede16666cd9ba038fb96eccc4e2fc42)

apps/ops-func/src/adapters/functions/split-delete-messages.ts
Garma00: [55869a0](https://github.com/pagopa/io-messages/pull/181/commits/55869a0249f8786161a8f1526966c7406e03865b)

apps/ops-func/src/adapters/functions/split-delete-messages.ts
Garma00: As you suggested we now use queue output, I think we don't need a separate function now, take a look here [55869a0](https://github.com/pagopa/io-messages/pull/181/commits/55869a0249f8786161a8f1526966c7406e03865b)

apps/ops-func/package.json
Garma00: [31a5bda](https://github.com/pagopa/io-messages/pull/181/commits/31a5bda3b57ba945e68a7a1f226b7afb2e70fd60)

apps/ops-func/src/adapters/functions/split-delete-messages.ts
```diff
@@ -1,29 +1,33 @@
...
-          "base64",
-        ),
-      );
+      return true;
     });
+
+    const serializedMessages = messages.map((message) => ({
+      fiscalCode: message.split(",")[0],
+      messageId: message.split(",")[1],
+    }));
```
lucacavallaro: I don't understand why you split this step

lucacavallaro (changes requested):

apps/ops-func/src/adapters/functions/split-delete-messages.ts
Garma00: So we can have an object `{fiscalCode: string, messageId: string}` as input for the function that will read from the queue.
At this stage `message` is a string `fiscalCode, messageId` @lucacavallaro

Garma00: [2349e7a](https://github.com/pagopa/io-messages/pull/181/commits/2349e7a051adecbfc373be85339ba8fd27cafd2b)

lucacavallaro (approved):

PR #184, opened 2025-03-21
Add fixtures

Garma00:
- Add new package fixtures
- modify etl-func so it can run locally using azurite for message-content
- add fixtures to etl-func
- Add new types on io-messages-common package so they can be shared between projects

Resolves IOCOM-2247


apps/etl-func/README.md
```diff
@@ -68,3 +68,13 @@ Expected output:
 ```bash
 it works!
 ```
+
+### Insert fixtures
+
+To insert fixtures simply run:
+
+```bash
+yarn fixtures:load <N>
```
lucamancosu: Seams that running **yarn build** on the workspace folder is required before execute the **yarn fixtures:load** command. If i'm right i propose to add this instruction

apps/etl-func/local.settings.json.example
```diff
@@ -6,6 +6,7 @@
     "FUNCTIONS_WORKER_RUNTIME": "node",
     "MESSAGE_CONTENT_STORAGE_URI": "https://testgiovannist.blob.core.windows.net/",
     "MESSAGE_CONTENT_CONTAINER_NAME": "messages",
+    "MESSAGE_CONTENT_STORAGE_CONNECTION_STRING":"DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://io-messages-azurite-1:10000/devstoreaccount1;QueueEndpoint=http://localhost:10001/devstoreaccount1;TableEndpoint=http://localhost:10002/devstoreaccount1;",
```
lucamancosu, line 9: probably we should replace io-messages-azurite-1  with localhost ?

packages/etl-fixtures/README.md
```diff
@@ -0,0 +1,13 @@
+# etl-fixtures
+
+This package contains fixtures for the etl function.
+If you want to test the etl function, you can use these fixtures by simply
+running the following command inside the `etl-func` app:
+
+```bash
+yarn fixtures:load
```
lucamancosu: Seams that running yarn build on the workspace folder is required before execute the yarn fixtures:load command. If i'm right i propose to add this instruction

apps/etl-func/README.md
```diff
@@ -68,3 +68,13 @@ Expected output:
 ```bash
 it works!
 ```
+
```
lucamancosu, line 71: there is probably a typo in the file

cp local.settings.json.example .local.settings.json

should be

**cp local.settings.json.example local.settings.json**

lucamancosu:
Should we create a script for the creation of the table for the tracking of the errors in the account storage ?

Garma00:
> Should we create a script for the creation of the table for the tracking of the errors in the account storage ?

@lucamancosu nice point! I guess we can add a `createIfNotExist` as we did for the blob container in the index file, I'll implement this, thanks!

apps/etl-func/README.md
Garma00: @lucamancosu running `yarn install` and `yarn build` in the root of the monorepo should be enough, I would avoid to add the instruction of building the single workspace just as we did with `io-messages-common`, do you agree?

packages/etl-fixtures/README.md
Garma00: https://github.com/pagopa/io-messages/pull/184#discussion_r2014080106

apps/etl-func/local.settings.json.example
Garma00: @lucamancosu `io-messages-azurite-1` is the name of the container, if we plan to make `devcontainer` as the standard way of testing the function locally I would leave it like that.
Anyway you made me notice that i left `localhost` in the last part of this connection string, if you agree with me I'll update it to use `io-messages-azurite-1`.

apps/etl-func/README.md
Garma00: @lucamancosu nice catch, I fixed it [e095708](https://github.com/pagopa/io-messages/pull/184/commits/e0957087619a5bf3223bd31448e24b12f78f2f61)

apps/etl-func/README.md
lucamancosu: Ok

apps/etl-func/local.settings.json.example
lucamancosu: Using the devcontainer we can reference to each service with the relative compose service. In the case of io-messages-azurite-1 we should use **azurite**. 
The string connection becomes like this

"DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://azurite:10000/devstoreaccount1;QueueEndpoint=http://azurite:10001/devstoreaccount1;TableEndpoint=http://azurite:10002/devstoreaccount1;",
(tell me if it works)

But this connection string will works only in a docker network(as the one in the devcontainer environment) where the compose service name is resolved.

I prefer to avoid the name of the container(io-messages-azurite-1) because it can change in some circumstances. 

What do you think ?

packages/etl-fixtures/src/adapters/cosmos/metadata-loader.ts
```diff
@@ -0,0 +1,69 @@
...
+   *
+   * @param count - The number of MessageMetadata objects to generate. Must be at least 1.
+   * @returns An array of MessageMetadata objects.
+   */
+  generateMany(count: number): MessageMetadata[] {
+    return Array.from({ length: count }, () => this.generate());
+  }
+
+  async load(metadata: MessageMetadata[]): Promise<void> {
+    Promise.all(metadata.map((m) => this.container.items.create(m)));
```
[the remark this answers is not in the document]
Garma00: [Here](https://github.com/pagopa/io-messages/pull/184/commits/188f4a6c7fe6a0c06e036c3e8ecb7f25770c5e37) I now return the Promise.all result so the caller can wait for the completion using await.
I don't want to use await here because I believe it should be up to the caller to decide whether to wait or not.

apps/etl-func/README.md
lucamancosu: It seams i cant start the fixture process because i'm missing FIXTURES_INCLUDE_PAYMENTS and FIXTURES_INCLUDE_REMOTE_CONTENT. 
I think this variables are not mandatory and they must have a default value. 
If there is a reason to make them mandatory then i suggest to add them to the local.settings.example

lucamancosu (approved):

apps/etl-func/README.md
Garma00: @lucamancosu those values should not be mandatory, they already have a default value take a look [here](https://github.com/pagopa/io-messages/blob/feats/add-etl-fixtures/packages/etl-fixtures/src/adapters/config.ts#L7)

lucamancosu: Still can't start the fixtures script without declaring FIXTURES_INCLUDE_PAYMENTS and FIXTURES_INCLUDE_REMOTE_CONTENT on local.settings

packages/fixtures/src/adapters/cosmos/metadata-loader.ts
```diff
@@ -0,0 +1,74 @@
...
+   * @returns An array of MessageMetadata objects.
+   */
+  generateMany(count: number): MessageMetadata[] {
+    return Array.from({ length: count }, () => this.generate());
+  }
+
+  async load(
+    metadata: MessageMetadata[],
+  ): Promise<ItemResponse<MessageMetadata>[]> {
+    const r = Promise.all(metadata.map((m) => this.container.items.create(m)));
```
lucamancosu: I think we must await there.

Garma00: I think it should be up to the caller weather to wait fot this load operation or not, we are here returning the Promise.all result so the caller can wait for it, in fact in the use case we have an `await`

apps/etl-func/package.json
```diff
@@ -9,6 +9,7 @@
   "scripts": {
     "build": "esbuild --bundle ./src/func.ts --format=esm --platform=node --target=node20 --outdir=dist --packages=external",
     "build:watch": "yarn build --watch",
+    "fixtures": "tsx scripts/fixtures.ts",
```
lucacavallaro: we can just use `js` and avoid `tsx`, what you think?

apps/etl-func/scripts/fixtures.ts
```diff
@@ -0,0 +1,42 @@
...
+const aadCredentials = new DefaultAzureCredential();
+
+if (existsSync(etlFunctionPath)) {
+  const settings = JSON.parse(readFileSync(etlFunctionPath, "utf8")).Values;
+  Object.keys(settings).forEach((key) => {
+    process.env[key] = settings[key];
+  });
+} else {
+  throw new Error("ETL function not found");
+}
```
lucacavallaro: why are you loading these variables onto `environment` when on line 30 you use them as variables? you can skip this part

packages/fixtures/src/adapters/blob/content-loader.ts
```diff
@@ -0,0 +1,60 @@
...
+            has_attachments: true,
+            has_remote_content: false,
+            id: ulid(),
+          }
+        : undefined,
+    }));
+  }
+
+  async load(messageContents: MessageContentWithIds[]): Promise<void> {
+    Promise.all(
```
lucacavallaro: you are returning the wrong promise here :)

packages/fixtures/src/adapters/config.ts
```diff
@@ -0,0 +1,21 @@
...
+  .toLowerCase()
+  .transform((x) => x === "true")
+  .pipe(z.boolean().default(false));
+
+export const loadFixturesOptionsSchema = z.object({
+  cosmosDatabaseName: z.string().min(1),
+  cosmosEndpoint: z.string().url(),
+  cosmosMessageContainerName: z.string().min(1),
+  includePayments: booleanFromString,
+  includeRemoteContents: booleanFromString,
```
lucacavallaro: why `strings` and not `boolean`?

packages/io-messages-common/src/types/log.ts
```diff
@@ -0,0 +1,5 @@
+export interface Logger {
+  error: (message: string) => void;
+  info: (message: string) => void;
+  log: (message: string) => void;
+}
```
lucacavallaro: is this abstraction really needed?

lucacavallaro (changes requested):

apps/etl-func/scripts/fixtures.ts
Garma00: @lucacavallaro Those variables are not accessible from process.env unless we load them from local.settings.json. Is there a cleaner way to do this?

packages/fixtures/src/adapters/config.ts
Garma00: From process.env we have to parse the string "true" into a boolean so I creted this utility

packages/io-messages-common/src/types/log.ts
Garma00: We can omit this and use the pino Logger as we did [here](https://github.com/pagopa/io-messages/blob/main/apps/etl-func/src/adapters/message.ts#L26C13-L26C19) for example, on the other hand I think that this approach goes against the port and adapter principle, what do you thinK @lucacavallaro?

packages/fixtures/src/adapters/blob/content-loader.ts
Garma00: @lucacavallaro You're right — I was reviewing this method again and I noticed a few things:
1. We're returning void here
2. In the load method for metadata, we're returning a type imported from the Cosmos SDK

I think a cleaner way to write this method would be to define a custom type like:
```ts
{ success: boolean; error?: Error }
```
This way, we can establish a clear 'port' between the use case and the adapter. Take a look at [05c80a5](https://github.com/pagopa/io-messages/pull/184/commits/05c80a53d92b226e2280add1349d3092dd771edf), do you agree?

apps/etl-func/package.json
Garma00: Well, good idea, did it here [3532cd4](https://github.com/pagopa/io-messages/pull/184/commits/3532cd46a662a481556d1f82043b722b06388bd5)

apps/etl-func/scripts/fixtures.ts
lucacavallaro: no, what I meant is that maybe we don't need to load these variable into environment

you can simply parse them to `settings` const and then pass to `loadFixturesOptionsSchema.parse` as `settings.MY_VAR`

packages/fixtures/src/adapters/blob/content-loader.ts
lucacavallaro: how is this interface different than returning `void` if OK and throwing `Error` if not?

packages/io-messages-common/src/types/log.ts
lucacavallaro: this is a little utility, we can relax this kind of constraints and avoid unnecessary abstractions

apps/etl-func/scripts/fixtures.ts
Garma00: Ok sorry now I understand, please take a look here [7fbba1e](https://github.com/pagopa/io-messages/pull/184/commits/7fbba1e3c9637bc654dd62ebfe3c3f20fe18978e)

packages/fixtures/src/adapters/config.ts
Garma00: @lucacavallaro since I made the changes you suggested [here](https://github.com/pagopa/io-messages/pull/184#discussion_r2042392557) now we can simply use boolean, take a look here [43cb5ee](https://github.com/pagopa/io-messages/pull/184/commits/43cb5eea71c37eff060f33f2566c25e9cc3c4a1e)

packages/io-messages-common/src/types/log.ts
Garma00: [032e652](https://github.com/pagopa/io-messages/pull/184/commits/032e652d666e877949346d834ed900def1001ca0)

packages/fixtures/src/adapters/blob/content-loader.ts
Garma00: Ok I agree, I simplified the code [ab56815](https://github.com/pagopa/io-messages/pull/184/commits/ab5681588687a2e8aa3f5c0aa414eefb5cbc0c0d)

packages/fixtures/src/adapters/blob/content-loader.ts
```diff
@@ -48,27 +46,15 @@ export class BlobContentLoader implements ContentLoader {
...
-    } catch (error) {
-      return {
-        error: new Error(
-          `Something went wrong trying to load ${messageContents.length} message contents | ${error}`,
-        ),
-        success: false,
-      };
-    }
+  async load(messageContents: MessageContentWithIds[]): Promise<void> {
+    await Promise.all(
```
lucacavallaro:
```suggestion
    return Promise.all(
```

packages/fixtures/src/adapters/cosmos/metadata-loader.ts
```diff
@@ -65,19 +63,7 @@ export class CosmosMetadataLoader implements MetadataLoader {
...
-    } catch (error) {
-      return {
-        error: new Error(
-          `Something went wrong trying to load ${metadata.length} message metadata | ${error}`,
-        ),
-        success: false,
-      };
-    }
+  async load(metadata: MessageMetadata[]): Promise<void> {
+    await Promise.all(metadata.map((m) => this.container.items.create(m)));
```
lucacavallaro:
```suggestion
    return Promise.all(metadata.map((m) => this.container.items.create(m)));
```

apps/etl-func/scripts/fixtures.js
```diff
@@ -18,25 +18,24 @@ const etlFunctionPath = resolve(__dirname, "../local.settings.json");
...
-  const settings = JSON.parse(readFileSync(etlFunctionPath, "utf8")).Values;
-  Object.keys(settings).forEach((key) => {
-    process.env[key] = settings[key];
-  });
-} else {
-  throw new Error("ETL function not found");
-}
+const getSettings = (settingsPath) => {
+  if (existsSync(settingsPath)) {
+    return JSON.parse(readFileSync(etlFunctionPath, "utf8")).Values;
```
lucacavallaro:
```suggestion
    return JSON.parse(readFileSync(settingsPath, "utf8")).Values;
```

apps/etl-func/scripts/fixtures.js
```diff
@@ -18,25 +18,24 @@ const etlFunctionPath = resolve(__dirname, "../local.settings.json");
...
+  cosmosEndpoint: settings.COMMON_COSMOS__accountEndpoint,
+  cosmosDatabaseName: settings.COMMON_COSMOS_DBNAME,
+  cosmosMessageContainerName: settings.COMMON_COSMOS_MESSAGES_CONTAINER_NAME,
+  storageConnectionString: settings.MESSAGE_CONTENT_STORAGE_CONNECTION_STRING,
+  storageMessageContentContainerName: settings.MESSAGE_CONTENT_CONTAINER_NAME,
+  includePayments: settings.FIXTURES_INCLUDE_PAYMENTS,
+  includeRemoteContents: settings.FIXTURES_INCLUDE_REMOTE_CONTENT,
 });
 
 loadFixtures(fixturesToGenerate, { ...options, aadCredentials });
```
lucacavallaro: why we don't `await` here?

lucacavallaro (changes requested):

packages/fixtures/src/adapters/blob/content-loader.ts
Garma00: Sorry I don't get the point, I agree that it should be better to return the promise so the caller can decide weather to wait or not but [here](https://github.com/pagopa/io-messages/pull/184#discussion_r2042394850) we said to return void

Garma00:
@lucacavallaro I made a refactor in order to respect clean architecture principles as we discussed

Garma00:
This feature is no longer needed.

PR #188, opened 2025-04-04
Move resources to com resource group

lucacavallaro:
Create new resources for: `repository`, `sending`, `redis` to `com` resource group.

In the next change we will remove the old ones and change the pointments.


Garma00:
@lucacavallaro it seems like the [plan](https://github.com/pagopa/io-messages/actions/runs/14260505531/job/39970849616?pr=188#step:8:362) is failing cause we are trying to add some roles to a function that does not exist already, we could add those roles in a separate PR, what's your take?

PR #190, opened 2025-04-09
Add checkout in labeler workflow

Garma00:


lucacavallaro (approved):
I pinned these actions to their latest commit

PR #192, opened 2025-04-11
Enable linting on legacy workspaces

lucamancosu:
Add working lint and lint:check scripts to sending-func and citizen-func


apps/citizen-func/.eslintrc.js
```diff
@@ -1,27 +1,26 @@
...
+    browser: true,
+    es6: true,
+    node: true,
+  },
+  ignorePatterns: [
+    "node_modules",
+    "generated",
+    "**/__tests__/*",
+    "**/__mocks__/*",
+    "**/__integrations__/*",
```
Garma00: we no longer use this directory inside `apps/` so we can remove this rule

apps/citizen-func/.eslintrc.js
```diff
@@ -1,27 +1,26 @@
...
+  parser: "@typescript-eslint/parser",
+  parserOptions: {
+    project: "tsconfig.json",
+    sourceType: "module",
+  },
+  extends: ["@pagopa/eslint-config/strong"],
+  rules: {
+    "no-invalid-this": "off",
+    "@typescript-eslint/no-invalid-this": "off",
+  },
```
Garma00, line 24: Are those needed? It was an empty block in the configuration before

apps/citizen-func/package.json
```diff
@@ -19,13 +20,26 @@
...
+    "eslint-plugin-prettier": "4.2.1",
+    "eslint-plugin-react": "7.32.2",
+    "eslint-plugin-sonarjs": "0.5.0",
     "modclean": "^3.0.0-beta.1",
-    "prettier": "^1.18.2",
+    "prettier": "1.19.1",
     "rimraf": "^6.0.1",
     "shx": "^0.3.4",
-    "typescript": "^4.3.5",
+    "typescript": "4.9.5",
```
Garma00, line 42: Why did you fix the version here?

apps/citizen-func/package.json
```diff
@@ -19,13 +20,26 @@
...
+    "eslint-plugin-functional": "2.0.0",
+    "eslint-plugin-import": "2.27.5",
+    "eslint-plugin-jsdoc": "30.7.13",
+    "eslint-plugin-prefer-arrow": "1.2.3",
+    "eslint-plugin-prettier": "4.2.1",
+    "eslint-plugin-react": "7.32.2",
+    "eslint-plugin-sonarjs": "0.5.0",
     "modclean": "^3.0.0-beta.1",
-    "prettier": "^1.18.2",
+    "prettier": "1.19.1",
```
Garma00, line 39: Why did you fix the version here?

apps/sending-func/package.json
```diff
@@ -26,13 +27,26 @@
...
+    "eslint-plugin-prefer-arrow": "1.2.3",
+    "eslint-plugin-prettier": "4.2.1",
+    "eslint-plugin-react": "7.32.2",
+    "eslint-plugin-sonarjs": "0.5.0",
     "npm-run-all": "^4.1.5",
     "prettier": "^1.18.2",
     "rimraf": "^6.0.1",
     "shx": "^0.3.4",
-    "typescript": "^4.3.5",
+    "typescript": "4.9.5",
```
Garma00, line 49: ditto

Garma00 (changes requested):

apps/citizen-func/.eslintrc.js
lucamancosu: Yes i think we need it. I've added this one because inside this file apps/citizen-func/utils/redis.ts the no-invalid-this rule is affecting a class even if it shouldn't.

apps/citizen-func/package.json
lucamancosu: I've fixed the same version of each package that comes from the old repo. If we change one of this packages versions that it can cause new errors while executing the lint check(errors that not exists on the original repo). I've used the versions from the yarn.lock file of the old repo.

apps/citizen-func/package.json
lucamancosu: I've fixed the same version of each package that comes from the old repo. If we change one of this packages versions that it can cause new errors while executing the lint check(errors that not exists on the original repo). I've used the versions from the yarn.lock file of the old repo.

apps/sending-func/package.json
lucamancosu: I've fixed the same version of each package that comes from the old repo. If we change one of this packages versions that it can cause new errors while executing the lint check(errors that not exists on the original repo). I've used the versions from the yarn.lock file of the old repo.

apps/citizen-func/.eslintrc.js
lucamancosu: 00b1388

apps/sending-func/package.json
lucacavallaro: Weren't these dependencies indirectly included via `@pagopa/eslint-config`?

lucamancosu: Yes they were while using node modules but pnp doesn't resolves it automatically or resolves some with different versions that cause what i've described above.

lucacavallaro: Ok, let me try in my environment!

lucacavallaro: Ok for me

lucacavallaro (approved):

Garma00 (approved):

PR #213, opened 2025-04-23
Add log information in health use-case

Garma00:


apps/ops-func/src/adapters/functions/health.ts
```diff
@@ -3,5 +3,16 @@ import { HttpHandler } from "@azure/functions";
...
+    if (failures.length > 0) {
+      return {
+        body: `Service connection failed, cause: ${failures}`,
+        status: 500,
+      };
+    }
+    return {
+      body: "It works!",
+      status: 200,
+    };
```
lucacavallaro: I would replace these responses, with a `JSON` like https://github.com/pagopa/io-fims/blob/main/apps/op-app/src/adapters/express/routes/health.ts#L9

lucacavallaro (changes requested):

apps/ops-func/src/adapters/functions/health.ts
Garma00: @lucacavallaro you mean something like this?
```ts
const failures = await healthUseCase.execute();
return {
  failures,
  status: failures.length > 0 ? 500 : 200,
  body: failures.length > 0 ? "Service connection failed" : "It works!", 
};
```

lucacavallaro:
```js
const failures = await healthUseCase.execute();
return {
  jsonBody: failures,
  status: failures.length > 0 ? 500 : 200
}
```

Garma00: [ee0f4ca](https://github.com/pagopa/io-messages/pull/213/commits/ee0f4ca516dccbdc467d129a39feb53dfd63e6eb)

lucacavallaro (approved):

PR #233, opened 2025-04-29
Refactor sending-func

lucacavallaro:
* Move source code into `src` folder
* Update toolchain to use the same scripts as modern workspaces
* Removed unused files and code
* Add `posthog`, `mockoon` for mocking e-mail and `io-backend`
* Add `bruno` requestes
* 🎉 this workspace works on our local environment


apps/sending-func/README.md
Garma00, line 1: Can we update the readme with newest steps to do to run this function locally? Also i think we can remove the section about integration tests

apps/sending-func/package.json
```diff
@@ -2,34 +2,37 @@
...
   },
   "devDependencies": {
     "@azure/functions": "^3.2.0",
     "@pagopa/eslint-config": "^4.0.3",
     "@pagopa/openapi-codegen-ts": "^14.0.0",
     "@types/express": "^4.17.11",
-    "@types/html-to-text": "^1.4.31",
     "@types/node": "^20.17.30",
-    "@types/nodemailer": "^4.6.8",
+    "@types/node-fetch": "^2.5.7",
```
Garma00, line 33: In future we could remove this dependency in order to use the standard fetch

Garma00 (changes requested):
@lucacavallaro can you update the local.settings.json?

apps/sending-func/README.md
lucacavallaro: I removed the README file as no workspace should require special instructions to get started. 

Just run `yarn start`

apps/sending-func/package.json
lucacavallaro: Yes, we will remove `node-fetch` as part of a future monorepo-wide update but that’s out of scope for now. This refactor only impacted dev dependencies and type-level source code.

Garma00 (approved):

PR #234, opened 2025-05-05
Remove unused code

lucamancosu:
Remove unused feature flags.

Resolves IOCOM-2307


apps/pushnotif-func/README.md
Garma00, line 1: We can remove the readme as suggested [here](https://github.com/pagopa/io-messages/pull/233#discussion_r2068771545)

Garma00 (approved):
GJ, just a small request 😄

apps/pushnotif-func/README.md
lucamancosu: If we remove the README i think there are some missing instructions for example the creation of the local.settings.json before run the function. What do you think ?

Garma00: "The creation of this file should be assumed when running any Function App in this monorepo. Do you agree?

lucacavallaro (approved):

PR #249, opened 2025-05-20
Remove citizen eslint disable

lucamancosu:
Resolves IOCOM-2366


lucamancosu:
I just left the eslint-disable inside the apps/citizen-func/src/utils/redis.ts.
Probably here a logger is needed but if i remember well this part must be centralized during the task for code duplication removal.
So i'm avoiding to introduce a logger now.
What do you think @lucacavallaro ?

apps/citizen-func/src/functions/GetMessages/__tests__/handler.test.ts
```diff
@@ -275,7 +275,7 @@ const getTaskMock = vi.fn(
 );
 vi.spyOn(redis, "getTask").mockImplementation(getTaskMock);
 
-const redisClientMock = {} as any;
+const redisClientMock = {} as unknown as RedisClientFactory;
```
lucacavallaro: can you replace this `as unknown as RedisClientFactory` with a proper `vi.mock` usage? we should avoid using `as` because is as unsafe as `any`

lucacavallaro (changes requested):

apps/citizen-func/src/functions/GetMessages/__tests__/handler.test.ts
lucamancosu: 02ca068e534842470bad0582495361baef8dbc0d

Garma00 (approved):

lucacavallaro (approved):

PR #252, opened 2025-05-22
Use config instead of env in initTelemetryClient

Garma00:


Garma00:
This way we can avoid the annoying `undefined` telemetry client

lucacavallaro (approved):

PR #255, opened 2025-05-27
 Move cosmos trigger Function from cqrs to sending

Garma00:
Depends on #252
Closes IOCOM-2244


lucacavallaro (approved):

apps/sending-func/src/functions/CosmosRemoteContentMessageConfigurationChangeFeed/index.ts
```diff
@@ -1,35 +1,33 @@
...
 const run = async (
   context: Context,
-  documents: ReadonlyArray<unknown>
+  documents: readonly unknown[],
 ): Promise<void> => {
   await handleRemoteContentMessageConfigurationChange(
     context,
     userRCConfigurationModel,
     telemetryClient,
-    config.MESSAGE_CONFIGURATION_CHANGE_FEED_START_TIME
```
lucacavallaro, line 32: why have you replaced this value?

apps/sending-func/CosmosRemoteContentMessageConfigurationChangeFeed/function.json
```diff
@@ -6,11 +6,11 @@
       "name": "cosmosRemoteContentMessageConfigurationTrigger",
       "direction": "in",
       "connection": "REMOTE_CONTENT_COSMOSDB",
-      "databaseName": "%REMOTE_CONTENT_COSMOSDB_NAME%",
+      "databaseName": "remote-content-cosmos-01",
```
lucacavallaro, line 9: this value has been always hardcoded?

apps/sending-func/src/functions/CosmosRemoteContentMessageConfigurationChangeFeed/index.ts
Garma00: @lucacavallaro This value was already set to '0'. I don't think we'll ever need to change it, so I think it's better to hardcode it

apps/sending-func/CosmosRemoteContentMessageConfigurationChangeFeed/function.json
Garma00: No, but I was thinking the same thing https://github.com/pagopa/io-messages/pull/255#discussion_r2149577892

lucacavallaro: this value should be parametized (the fact that cosmos has a `-01` is an hint!), since it can different in different environment (local machine, cosmos development) and the database can be migrated in the future for any reason

PR #257, opened 2025-05-28
Use session-manager client to resolve the session in Notify function

Garma00:
Depends on #260 
Resolves IOCOM-2381


apps/sending-func/src/functions/Notify/index.ts
```diff
@@ -86,20 +86,25 @@ const profileModel = new ProfileModel(
...
+    }),
 });
 
 // Add express route
 app.post(
   "/api/v1/notify",
   Notify(
     getUserProfileReader(profileModel),
-    getUserSessionStatusReader(sessionClient),
+    getUserSessionStatusReader(
```
lucacavallaro: my curiosity: doesn't the session manager replace also the user profile reader?

infra/resources/_modules/function_app_sending/locals.tf
```diff
@@ -34,6 +34,10 @@ locals {
       BACKEND_BASE_URL = "https://io-p-app-appbackendli.azurewebsites.net"
       BACKEND_TOKEN    = var.appbackendli_token
 
+      // TODO: get api key from kv
+      SESSION_MANAGER_API_KEY  = "",
+      SESSION_MANAGER_BASE_URL = var.session_manager_base_url,
```
lucacavallaro: !! don't forget to resolve this todo

lucacavallaro (changes requested):

apps/sending-func/src/functions/Notify/index.ts
Garma00: Unfortunately, no. The getUserSessionStatusReader only returns a boolean indicating whether the user has an active session. We still need to access the user preferences to check notification settings, see [here](https://github.com/pagopa/io-messages/blob/refactors/call-sm-instead-of-be/apps/sending-func/src/functions/Notify/handler.ts#L100-L100) for example.

infra/resources/_modules/function_app_sending/locals.tf
Garma00: Sure, we first need to merge #260 😄

lucacavallaro (approved):

PR #265, opened 2025-06-03
Avoid creating new NH client for every operation

Garma00:
- Create a single nh client when starting the function using a factory
- Remove calls to legacy NH (no more used)

Resolves IOCOM-2400

### DIFF
--- a/apps/pushnotif-func/utils/notificationhubServicePartition.ts
+++ b/apps/pushnotif-func/utils/notificationhubServicePartition.ts
@@ -8,7 +8,7 @@ import { InstallationId } from "../generated/notifications/InstallationId";
 import { IConfig } from "./config";
 import { DisjoitedNotificationHubPartitionArray } from "./types";
 
-export const NotificationHubConfig = t.interface({
+export const NotificationHubConfig = t.type({
   AZURE_NH_ENDPOINT: NonEmptyString,
   AZURE_NH_HUB_NAME: NonEmptyString,
 });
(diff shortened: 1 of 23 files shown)


apps/pushnotif-func/utils/notificationhubServicePartition.ts
```diff
@@ -67,3 +68,59 @@ export const buildNHClient = ({
   AZURE_NH_HUB_NAME,
 }: NotificationHubConfig): NotificationHubsClient =>
   new NotificationHubsClient(AZURE_NH_ENDPOINT, AZURE_NH_HUB_NAME);
+
+export class NotificationHubPartitionFactory {
+  #notificationHubPartitions: DisjoitedNotificationHubPartitionArray;
+  #partitionFour: NotificationHubsClient;
+  #partitionOne: NotificationHubsClient;
+  #partitionThree: NotificationHubsClient;
+  #partitionTwo: NotificationHubsClient;
```
lucacavallaro: ugh

can't you use a map instead of four different variables?

apps/pushnotif-func/utils/notificationhubServicePartition.ts
```diff
@@ -67,3 +68,59 @@ export const buildNHClient = ({
...
+      this.#notificationHubPartitions[1].endpoint,
+      this.#notificationHubPartitions[1].name,
+    );
+    this.#partitionThree = new NotificationHubsClient(
+      this.#notificationHubPartitions[2].endpoint,
+      this.#notificationHubPartitions[2].name,
+    );
+    this.#partitionFour = new NotificationHubsClient(
+      this.#notificationHubPartitions[3].endpoint,
+      this.#notificationHubPartitions[3].name,
```
lucacavallaro: this code does not scale :| we have for loops, dynamic structures...

lucacavallaro (changes requested):

apps/pushnotif-func/utils/notificationhubServicePartition.ts
```diff
@@ -67,3 +68,59 @@ export const buildNHClient = ({
   AZURE_NH_HUB_NAME,
 }: NotificationHubConfig): NotificationHubsClient =>
   new NotificationHubsClient(AZURE_NH_ENDPOINT, AZURE_NH_HUB_NAME);
+
+export class NotificationHubPartitionFactory {
+  #notificationHubPartitions: DisjoitedNotificationHubPartitionArray;
```
gunzip:
```suggestion
  #notificationHubPartitions: DisjointedNotificationHubPartitionArray;
```

apps/pushnotif-func/utils/notificationhubServicePartition.ts
Garma00: I opted for a `Record` cause with a Map we have to deal with the `undefined` case when using the `get` method, please take a look [d0521ed](https://github.com/pagopa/io-messages/pull/265/commits/d0521eda9c4dda6d888ab59afc758b22e3ff2714)

apps/pushnotif-func/utils/notificationhubServicePartition.ts
Garma00: [d0521ed](https://github.com/pagopa/io-messages/pull/265/commits/d0521eda9c4dda6d888ab59afc758b22e3ff2714)

apps/pushnotif-func/utils/notificationhubServicePartition.ts
Garma00: This is an already defined type, I think it is better to change this type name in a separate PR cause it is used in several files, thanks tho!

lucacavallaro (approved):

PR #273, opened 2025-06-05
Use new redis on com services

lucamancosu:


infra/resources/_modules/web_apps/remote-content-func.tf
lucacavallaro, line 1: @lucamancosu it seems that you are starting from the wrong branch, this file should not be here

lucacavallaro (changes requested):

infra/resources/_modules/web_apps/remote-content-func.tf
lucamancosu: Yes my fault. Can we archive/delete this PR ?

PR #276, opened 2025-06-05
Add sending-func kv permissions

Garma00:


infra/resources/prod/main.tf
```diff
@@ -39,7 +39,7 @@ module "functions_messages_sending" {
   domain              = "msgs"
   resource_group_name = data.azurerm_resource_group.itn_messages.name
 
-  com_kv_name = module.key_vaults.kv_name
+  com_kv = module.key_vaults.com
```
lucacavallaro: I think we should make this input variable generic

```suggestion
  key_vault = module.key_vaults.com
```

lucacavallaro (changes requested):

infra/resources/prod/main.tf
Garma00: [2a5b0ee](https://github.com/pagopa/io-messages/pull/276/commits/2a5b0eeb01b9c27998e2ceefe0d833c7921b9729)

lucacavallaro (approved):

PR #277, opened 2025-06-05
Remove APPINSIGHTS_DISABLE from config

Garma00:


lucacavallaro (changes requested):
can you apply this change also in the other workspaces?

lucacavallaro (approved):

PR #279, opened 2025-06-06
Remove files and tables from healthcheck

Garma00:


Garma00:
We do not use those services, also files is not supported from azurite, this make the `Info` function fail locally.

lucacavallaro (approved):

PR #288, opened 2025-06-09
Fix cqrs func app settings

acuiuli:
- remove unused env var from config
- add required env vars to infra
- new message-paymentupdater-failures queue created in domain storage


apps/cqrs-func/local.settings.json.example
```diff
@@ -2,7 +2,6 @@
   "IsEncrypted": false,
   "Values": {
     "FUNCTIONS_WORKER_RUNTIME": "node",
-    "AzureWebJobsStorage": "{AzureWebJobsStorage}",
```
lucacavallaro, line 5: this setting is mandatory (it should point to the internal storage used by the function), but should not referenced anywhere in the application code

infra/resources/_modules/web_apps/cqrs-func.tf
```diff
@@ -156,7 +159,7 @@ module "cqrs_func" {
 
   tags = var.tags
 
-  application_insights_connection_string   = var.application_insights.connection_string
```
lucacavallaro: why are you removing the connection string? the instrumentation key is deprecated and must not be used anymore

apps/cqrs-func/env.example
lucacavallaro, line 1: this .env file is still necessary? we've removed that kind of file from other legacy workspaces

lucacavallaro (changes requested):

lucacavallaro (approved):

PR #307, opened 2025-06-17
Change type on pushnotif package.json

lucamancosu:
Type=module was accidentally committed. It is causing a runtime error regarding some module exports


lucacavallaro:
@lucamancosu why is this change required?

lucamancosu:
> @lucamancosu why is this change required?

because with type=module we have a runtime error

lucacavallaro (approved):

PR #312, opened 2025-06-19
Import services-func

Garma00:
Create a new function `app services-func` with all the domain related functions that are placed in pagopa/io-functions-services at the moment:
- GetMessage
- CreateMessage
- ProcessMessage
- OnFailedProcessMessage
- CreateNotification
- EmailNotification
- WebhookNotification
- Info

Closes IOCOM-2416


apps/services-func/local.settings.json.example
```diff
@@ -0,0 +1,37 @@
...
+    "FUNCTIONS_WORKER_RUNTIME": "node",
+    "AzureWebJobsStorage": "DefaultEndpointsProtocol=https;AccountName=myaccount;AccountKey=mykey;EndpointSuffix=core.windows.net",
+    "APPINSIGHTS_INSTRUMENTATIONKEY": "aKey",
+    "MESSAGE_CONTAINER_NAME": "messages",
+    "COSMOSDB_NAME": "io-messages",
+    "COSMOSDB_KEY": "cosmosDBKey",
+    "COSMOSDB_URI": "https://io-d-itn-common-cosno-01.documents.azure.com:443/",
+    "WEBHOOK_CHANNEL_URL": "https://domain.com",
+    "MAIL_FROM": "IO - l’app dei servizi pubblici <no-reply@io.italia.it>",
+    "QueueStorageConnection": "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://azurite:10000/devstoreaccount1;QueueEndpoint=http://azurite:10001/devstoreaccount1;TableEndpoint=http://azurite:10002/devstoreaccount1;",
```
lucamancosu: I think this is not used. Can we remove it ?

apps/services-func/local.settings.json.example
```diff
@@ -0,0 +1,37 @@
...
+    "SUBSCRIPTIONS_FEED_TABLE": "SubscriptionsFeedByDay",
```
lucamancosu: I think this is not used. Can we remove it ?

infra/resources/_modules/web_apps/services-func.tf
```diff
@@ -0,0 +1,192 @@
...
+
+locals {
+  services_func = {
+    app_settings = {
+      FUNCTIONS_WORKER_RUNTIME       = "node"
+      WEBSITE_RUN_FROM_PACKAGE       = "1"
+      WEBSITE_DNS_SERVER             = "168.63.129.16"
+      FUNCTIONS_WORKER_PROCESS_COUNT = 4
+      NODE_ENV                       = "production"
+
```
lucamancosu, line 9: Are APPINSIGHTS_INSTRUMENTATIONKEY, INTERNAL_STORAGE_CONNECTION_STRING, FF_DISABLE_INCOMPLETE_SERVICES and MAILHOG_HOSTNAME variables required here ?

apps/services-func/local.settings.json.example
Garma00: Nice catch, please take a look [here](https://github.com/pagopa/io-messages/pull/312/commits/fde635f5fa863050f873b86479b774ff9c14084f)

apps/services-func/local.settings.json.example
Garma00: Nice catch, please take a look [here](https://github.com/pagopa/io-messages/pull/312/commits/fde635f5fa863050f873b86479b774ff9c14084f)

infra/resources/_modules/web_apps/services-func.tf
Garma00: I think that APPINSIGHTS_INSTRUMENTATIONKEY and INTERNAL_STORAGE_CONNECTION_STRING are automatically set.
FF_DISABLE_INCOMPLETE_SERVICES has a default in the config and we are going to use it.
MAILHOG_HOSTNAME is only needed during local dev as we can see [here](https://github.com/pagopa/io-functions-commons/blob/master/src/mailer/config.ts#L36-L36)

lucamancosu (approved):

PR #337, opened 2025-06-30
Point APIM to the new paymentup instance

lucacavallaro:


infra/resources/_modules/apim/payment_updater.tf
```diff
@@ -43,6 +43,14 @@ resource "azurerm_api_management_api" "payments_updater_api_v1" {
   }
 }
 
+resource "azurerm_api_management_named_value" "example" {
```
Garma00: Why "example" here?

Garma00 (changes requested):

infra/resources/_modules/apim/payment_updater.tf
lucacavallaro: Because I copied that from the doc! Fixed

Garma00 (approved):

PR #349, opened 2025-06-30
Import resources from io-infra

lucacavallaro:


Garma00:
@lucacavallaro  why is the plan failing?
Also, are we ok with the reminder cosmos collection being recreated?

lucacavallaro:
> @lucacavallaro why is the plan failing? Also, are we ok with the reminder cosmos collection being recreated?

@Garma00 we should import that collection, we don't want to recreate it, the plan is failing because it references a data (acr) that we removed yesterday

Garma00 (approved):

acuiuli (approved):

PR #384, opened 2025-07-17
Move notify api inside pushnotif-func

lucamancosu:
Resolves IOCOM-2483


infra/resources/_modules/web_apps/pushnotif-func.tf
```diff
@@ -125,6 +125,20 @@ locals {
       # Possible values : "none" | "all" | "beta" | "canary"
       NH_PARTITION_FEATURE_FLAG = "all"
 
+
+      MESSAGE_CONTAINER_NAME                    = "message-content"
+      MESSAGE_CONTENT_STORAGE_CONNECTION_STRING = var.message_content_storage.connection_string
```
Garma00, line 130: Now that this funciton need to access to:
- message content storage
- domain cosmos
- common cosmos

I would update the Info handler to check the connection to those new resources, what's your take? You can take the `sending-func` healthcheck as an example @lucamancosu

Garma00 (approved):

PR #392, opened 2025-07-23
Remove sending-func

lucamancosu:
Resolves IOCOM-2494


apps/reminder/pom.xml
Garma00, line 1: Do we need to make a new release of the reminder in this case? @lucamancosu @acuiuli
If so, we need to add a changeset

Garma00 (changes requested):

apps/reminder/pom.xml
acuiuli: if the openapi is the same but only the path is changed than a new release is not mandatory but it is better to do it anyway for keeping everything up to date.

Garma00 (approved):

PR #410, opened 2025-08-01
Add lollipop client in common package

acuiuli:
Add the lollipop client with the call to the generateLcParams api in the io-messages-common package.

Resolves [IOCOM-2524](https://pagopa.atlassian.net/browse/IOCOM-2524)

[IOCOM-2524]: https://pagopa.atlassian.net/browse/IOCOM-2524?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/io-messages-common/src/adapters/lollipop/__tests__/lollipop-client.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  it("should throw an error if fetch fails", async () => {
+    const returnedError = new Error("Network error");
+    const fetchSpy = vi
+      .spyOn(globalThis, "fetch")
+      .mockRejectedValueOnce(returnedError);
+
+    try {
+      await client.generateLCParams(anAssertionRef, aSignatureInput);
+      throw new Error("Expected error to be thrown");
```
Garma00: Why this throw here? the generateLCParams should throw the error
```suggestion
```

packages/io-messages-common/src/adapters/lollipop/__tests__/lollipop-client.test.ts
```diff
@@ -0,0 +1,104 @@
...
+  it("throws an error when the API responds with an error", async () => {
+    const fetchSpy = vi.spyOn(globalThis, "fetch").mockResolvedValueOnce({
+      json: async () => aProblemJson,
+      ok: false,
+      status: 400,
+    } as Response);
+
+    try {
+      await client.generateLCParams(anAssertionRef, aSignatureInput);
+      throw new Error("Expected error to be thrown");
```
Garma00: Ditto
```suggestion
```

packages/io-messages-common/src/adapters/lollipop/__tests__/lollipop-client.test.ts
```diff
@@ -0,0 +1,104 @@
...
+        headers: expect.objectContaining({
+          "X-Functions-Key": apiKey,
+          "content-type": "application/json",
+        }),
+        method: "POST",
+      }),
+    );
+  });
+
+  it("throws an error when the API responds with an error", async () => {
```
Garma00, line 42: In those 2 `it` we are not sure that the `expect` in the `catch` block will be executed, if for some reason the generateLCParams does not throw then the expect would be ignored.
I would suggest to use a different approach as we did [here](https://github.com/pagopa/io-messages/blob/main/apps/services-func/src/functions/OnFailedProcessMessage/__tests__/handler.test.ts#L129-L129).

This way we ensure that the generateLCParams failed with the expected error

Garma00 (changes requested):

lucamancosu (approved):

lucamancosu:
GJ! Why we have introduced the definitions folder ? Those definitions will be used outside the lollypop client ? 

My advice is to put all the necessary in a single file(the lollipop-client.ts file in this case)

packages/io-messages-common/src/adapters/lollipop/__tests__/lollipop-client.test.ts
acuiuli: You are right, done [here](https://github.com/pagopa/io-messages/pull/410/commits/55fc0fbd179379bd886e224feee8a2642ce61ae8)

packages/io-messages-common/src/adapters/lollipop/__tests__/lollipop-client.test.ts
acuiuli: [here](https://github.com/pagopa/io-messages/pull/410/commits/55fc0fbd179379bd886e224feee8a2642ce61ae8)

packages/io-messages-common/src/adapters/lollipop/__tests__/lollipop-client.test.ts
acuiuli: [here](https://github.com/pagopa/io-messages/pull/410/commits/55fc0fbd179379bd886e224feee8a2642ce61ae8)

acuiuli:
> GJ! Why we have introduced the definitions folder ? Those definitions will be used outside the lollypop client ?
> 
> My advice is to put all the necessary in a single file(the lollipop-client.ts file in this case)

The definitions will be used in the functions that uses this client for the response (`LcParams`) `AssertionRef` and `SignatureInput` at least, but probably also more are needed. That's why I put them in the definitions folders and divided them in single files

Garma00 (approved):

PR #416, opened 2025-08-27
Change message-status-event schema

lucamancosu:
Closes IOCOM-2614


Garma00 (approved):
LGTM

apps/etl-func/src/domain/__tests__/message-status-event.test.ts
robbcocco, line 1: The test `Given a valid MessageStatus it should return a valid MessageStatusEvent` seems like a duplicate of `Given a valid MessageStatus, when the version is 0 it should return a valid MessageStatusEvent with op CREATE` in `apps/etl-func/src/adapters/__tests__/message-status.test.ts`

Garma00 (approved):

apps/etl-func/src/domain/__tests__/message-status-event.test.ts
Garma00: Nice catch! [5bac2bc](https://github.com/pagopa/io-messages/pull/416/commits/5bac2bc49a7b595d6736049a419ee38f5c142103)

PR #424, opened 2025-09-19
Add SEND AAR QRCode and Get Notification API

lucamancosu:
Closes: IOCOM-2579


apps/send-func/package.json
```diff
@@ -9,7 +9,7 @@
   "scripts": {
     "build": "esbuild --bundle ./src/func.ts --format=esm --platform=node --target=node20 --outdir=dist --packages=external",
     "build:watch": "yarn build --watch",
-    "start": "func start",
+    "start": "func start --port 7074",
```
acuiuli: why do we start it with this port?

apps/send-func/src/adapters/functions/aar-notifications.ts
```diff
@@ -0,0 +1,87 @@
...
+        const problemJson = problemJsonSchema.parse(err.body);
+
+        return {
+          jsonBody: {
+            detail: "Internal server error",
+            errors: problemJson.errors,
+            status: err.status,
+          },
+          status: 500,
+        };
```
acuiuli, line 67:
```suggestion
        return {
          jsonBody: problemJson
          status: 500,
        };
```

apps/send-func/src/adapters/functions/aar-notifications.ts
```diff
@@ -0,0 +1,87 @@
...
+
+      const response = await client.getReceivedNotification(
+        iun,
+        sendHeaders,
+        mandateId,
+      );
+
+      return { jsonBody: response, status: 200 };
+    } catch (err) {
+      if (err instanceof Error && err.name === "ZodError") {
```
acuiuli: Why don't you manage all the parse errors one by one by using the `safeParse`? Like this we are mixing all the parsing error and the Zod default error message is not really good to send back to the app, I think it would be better to customize every error individually. Also by doing it like that we can manage all the Error and just use the message instead of managing only the ZodError and losing the error details when it is not a ZodError

apps/send-func/src/adapters/functions/aar-notifications.ts
```diff
@@ -0,0 +1,87 @@
...
+  ): HttpHandler =>
+  async (
+    request: HttpRequest,
+    context: InvocationContext,
+  ): Promise<HttpResponseInit> => {
+    try {
+      const isTest = request.query.get("isTest") === "true";
+      const sendHeaders = sendHeadersSchema.parse(
+        context.extraInputs.get(lollipopExtraInputsCtxKey),
+      );
```
acuiuli: Shouldn't this fail everytime? The lollipop middleware return a `LollipopHeaders` and not a `SendHeaders`. You should first parse the `LollipoHeaders` and then add the `x-pagopa-cx-taxid` to create a `SendHeaders`

apps/send-func/package.json
lucamancosu: 0c5a0adcb89c695c8a40684a9f949cea4f603891

apps/send-func/src/adapters/functions/aar-notifications.ts
lucamancosu: d80ef63a1ddf88eec5753ad67007cd365496b2b5

apps/send-func/src/adapters/functions/aar-notifications.ts
lucamancosu: d80ef63a1ddf88eec5753ad67007cd365496b2b5

apps/send-func/src/adapters/functions/aar-notifications.ts
lucamancosu: d80ef63a1ddf88eec5753ad67007cd365496b2b5

apps/send-func/local.settings.json.example
```diff
@@ -6,6 +6,8 @@
     "NOTIFICATION_CLIENT_API_KEY": "anApiKey",
     "NOTIFICATION_CLIENT_BASE_URL": "https://produrl.com",
     "NOTIFICATION_CLIENT_UAT_API_KEY": "anApiKey",
-    "NOTIFICATION_CLIENT_UAT_BASE_URL": "https://uaturl.com"
+    "NOTIFICATION_CLIENT_UAT_BASE_URL": "https://uaturl.com",
+    "LOLLIPOP_API_BASE_URL": "https://io-p-itn-auth-lollipop-func-02.azurewebsites.net/api/v1",
```
Garma00, line 10: Can we use a mock for this?

apps/send-func/src/adapters/functions/aar-notifications.ts
```diff
@@ -0,0 +1,99 @@
...
+export const getNotification =
+  (
+    notificationClient: NotificationClient,
+    uatNotificationClient: NotificationClient,
+  ): HttpHandler =>
+  async (
+    request: HttpRequest,
+    context: InvocationContext,
+  ): Promise<HttpResponseInit> => {
+    try {
```
Garma00: Why is everything inside a try even thought we are using safeParse?

apps/send-func/src/adapters/functions/aar-notifications.ts
```diff
@@ -0,0 +1,99 @@
...
+      );
+
+      const sendHeaders = sendHeadersSchema.safeParse({
+        "x-pagopa-cx-taxid": request.headers.get("x-pagopa-cx-taxid"),
+        "x-pagopa-pn-io-src":
+          request.headers.get("x-pagopa-pn-io-src") || undefined,
+        ...lollipopHeaders.data,
+      });
+
+      if (!lollipopHeaders.success || !sendHeaders.success) {
```
Garma00:
```suggestion
      if (!sendHeaders.success) {
```

There is no need to check for lollipopHeaders.success here, sendHeaders.success would be false if lollipopHeaders.success is

apps/send-func/src/adapters/functions/aar-notifications.ts
```diff
@@ -0,0 +1,99 @@
...
+        return {
+          jsonBody: {
+            detail: "Malformed headers",
+            status: 400,
```
Garma00: It could be a good idea to improve this error message, maybe we can use `sendHeaders.error` here

apps/send-func/src/adapters/functions/aar-notifications.ts
```diff
@@ -0,0 +1,99 @@
...
+
+      if (!iun.success || !mandateId.success) {
+        return {
+          jsonBody: {
+            detail: "Malformed request",
+            status: 400,
+          },
+          status: 400,
+        };
+      }
```
Garma00: The entire logic before this 400 could be extracted in a single function, something like `validateGetNotificationrequest`, this way this logic can be more testable using unit test.

Also, reading this code it is difficult to tell where the request validation begins/ends and where the BL begins

apps/send-func/src/adapters/functions/aar-notifications.ts
```diff
@@ -0,0 +1,99 @@
...
+
+      const iun = iunSchema.safeParse(request.params.iun);
+      const mandateId = request.query.has("mandateId")
+        ? mandateIdSchema.safeParse(request.query.get("mandateId"))
+        : { data: undefined, success: true };
+
+      if (!iun.success || !mandateId.success) {
+        return {
+          jsonBody: {
+            detail: "Malformed request",
```
Garma00: ditto

apps/send-func/src/adapters/functions/aar-qrcode-check.ts
```diff
@@ -0,0 +1,103 @@
...
+    notificationClient: NotificationClient,
+    uatNotificationClient: NotificationClient,
+  ): HttpHandler =>
+  async (
+    request: HttpRequest,
+    context: InvocationContext,
+  ): Promise<HttpResponseInit> => {
+    try {
+      const isTest = request.query.get("isTest") === "true";
+      const client = isTest ? uatNotificationClient : notificationClient;
```
Garma00: Can we extract this logic to a separate function? something like 
```ts
getSendClien(isTest: boolean | undefined): UatNotificationClient | NotificationClient
```
So we can reuse it

apps/send-func/src/adapters/functions/aar-qrcode-check.ts
```diff
@@ -0,0 +1,103 @@
...
+        context.extraInputs.get(lollipopExtraInputsCtxKey),
+      );
+      const sendHeaders = sendHeadersSchema.safeParse({
+        "x-pagopa-cx-taxid": request.headers.get("x-pagopa-cx-taxid"),
+        "x-pagopa-pn-io-src":
+          request.headers.get("x-pagopa-pn-io-src") || undefined,
+        ...lollipopHeaders.data,
+      });
+
+      if (!lollipopHeaders.success || !sendHeaders.success) {
```
Garma00: ditto

apps/send-func/src/adapters/functions/aar-qrcode-check.ts
```diff
@@ -0,0 +1,103 @@
...
+        return {
+          jsonBody: {
+            detail: "Malformed headers",
```
Garma00: ditto

apps/send-func/src/adapters/functions/aar-qrcode-check.ts
```diff
@@ -0,0 +1,103 @@
...
+export const aarQRCodeCheck =
+  (
+    notificationClient: NotificationClient,
+    uatNotificationClient: NotificationClient,
+  ): HttpHandler =>
+  async (
+    request: HttpRequest,
+    context: InvocationContext,
+  ): Promise<HttpResponseInit> => {
+    try {
```
Garma00: ditto

apps/send-func/src/adapters/functions/aar-qrcode-check.ts
```diff
@@ -0,0 +1,103 @@
...
+        return {
+          jsonBody: {
+            detail: "Malformed headers",
+            status: 400,
+          },
+          status: 400,
+        };
+      }
+
+      const rawBody = await request.json();
```
Garma00: How are we handle a throw here? If I am not mistaken this will cause a 500 with the current logic, would be better a `400` here?

apps/send-func/src/adapters/functions/aar-qrcode-check.ts
```diff
@@ -0,0 +1,103 @@
...
+      const parsedBody = checkQrMandateRequestSchema.safeParse(rawBody);
+      if (!parsedBody.success) {
+        return {
+          jsonBody: {
+            detail: "Malformed body",
```
Garma00:
```suggestion
            detail: `Malformed request body: ${parsedBody.error}`,
```

apps/send-func/src/adapters/functions/aar-qrcode-check.ts
```diff
@@ -0,0 +1,103 @@
...
+            status: 400,
+          },
+          status: 400,
```
Garma00: Can we avoid this duplication?

apps/send-func/src/adapters/functions/__tests__/aar-notifications.test.ts
```diff
@@ -0,0 +1,188 @@
...
+    });
+    request.query.set("isTest", "false");
+    request.headers.set("x-pagopa-cx-taxid", aFiscalCode);
+
+    await expect(handler(request, context)).resolves.toEqual({
+      jsonBody: aThirdPartyMessage,
+      status: 200,
+    });
+
+    const parseHeaders = sendHeadersSchema.parse(sendHeaders);
```
Garma00: You can parse the those headers during the instantiation instead of parsing for every test case

apps/send-func/src/adapters/functions/__tests__/aar-notifications.test.ts
```diff
@@ -0,0 +1,188 @@
...
+    expect(getReceivedNotificationSpy).toHaveBeenCalledWith(
+      aIun,
+      parseHeaders,
+      undefined,
+    );
+    expect(uatGetReceivedNotificationSpy).not.toHaveBeenCalledOnce();
```
Garma00:
```suggestion
    expect(uatGetReceivedNotificationSpy).not.toHaveBeenCalled();
```

apps/send-func/src/adapters/functions/__tests__/aar-notifications.test.ts
```diff
@@ -0,0 +1,188 @@
...
+    request.query.set("isTest", "true");
+    request.headers.set("x-pagopa-cx-taxid", aFiscalCode);
+
+    await expect(handler(request, context)).resolves.toEqual({
+      jsonBody: aThirdPartyMessage,
+      status: 200,
+    });
+
+    const parseHeaders = sendHeadersSchema.parse(sendHeaders);
+    expect(getReceivedNotificationSpy).not.toHaveBeenCalledOnce();
```
Garma00:
```suggestion
    expect(getReceivedNotificationSpy).not.toHaveBeenCalled();
```

apps/send-func/src/adapters/functions/__tests__/aar-notifications.test.ts
```diff
@@ -0,0 +1,188 @@
...
+      "x-pagopa-lollipop-assertion-type": anAssertionType,
+      "x-pagopa-lollipop-auth-jwt": "an auth jwt",
+      "x-pagopa-lollipop-original-method": anOriginalMethod,
+    };
+    const mandateId = crypto.randomUUID();
+    request.query.set("mandateId", mandateId);
+
+    context.extraInputs.set("lollipopHeaders", malformedLollipopHeaders);
+
+    const result2 = (await handler(request, context)) as HttpResponseInit;
```
Garma00: Why do we need this cast? We should be 100% sure that we are returning something with `jsonBody`, this is the value that we are returning as response

apps/send-func/src/adapters/functions/__tests__/aar-qrcode-check.test.ts
```diff
@@ -0,0 +1,206 @@
...
+      .spyOn(request, "json")
+      .mockResolvedValue({ aarQrCodeValue: anAarQrCodeValue });
+
+    await expect(handler(request, context)).resolves.toEqual({
+      jsonBody: aCheckQrMandateResponse,
+      status: 200,
+    });
+    expect(requestBodyJson).toHaveBeenCalledOnce();
+
+    const parseHeaders = sendHeadersSchema.parse(sendHeaders);
```
Garma00: ditto

apps/send-func/src/adapters/functions/__tests__/aar-qrcode-check.test.ts
```diff
@@ -0,0 +1,206 @@
...
+    expect(checkAarQrCodeIOSpy).toHaveBeenCalledWith(
+      anAarQrCodeValue,
+      parseHeaders,
+    );
+    expect(uatCheckAarQrCodeIOSpy).not.toHaveBeenCalledOnce();
```
Garma00:
```suggestion
    expect(uatCheckAarQrCodeIOSpy).not.toHaveBeenCalled();
```

apps/send-func/src/adapters/functions/__tests__/aar-qrcode-check.test.ts
```diff
@@ -0,0 +1,206 @@
...
+      .mockResolvedValue({ aarQrCodeValue: anAarQrCodeValue });
+
+    await expect(handler(request, context)).resolves.toEqual({
+      jsonBody: aCheckQrMandateResponse,
+      status: 200,
+    });
+    expect(requestBodyJson).toHaveBeenCalledOnce();
+
+    const parseHeaders = sendHeadersSchema.parse(sendHeaders);
+    expect(checkAarQrCodeIOSpy).not.toHaveBeenCalledOnce();
```
Garma00:
```suggestion
    expect(checkAarQrCodeIOSpy).not.toHaveBeenCalled();
```

Garma00 (changes requested):

apps/send-func/src/adapters/functions/aar-notifications.ts
lucamancosu: The idea was to not send the zod errors into the response. Are you ok with that ?
In this first phase we are simply going to use a generic malformed error i think.

What we can give a litle bit more of details like: Malformed lollipop errors, Malformed iun parameter ecc. But i don't know if is necessary to comunicate to the clients.

Garma00: We can discuss about it, I thin we would debug faster in case of errors by sending a more detailed error to the client, the client doesn't need to show to the user all the details we send cc @acuiuli @robbcocco

apps/send-func/src/adapters/functions/aar-notifications.ts
lucamancosu: My first implementation was like this.  All the validation logic was inside a function and this was making the handler easier to read.
I've change to the current implentation because the validation function had introduced the need to use additional data structures to return the data needed to call the client.checkAarQrCodeIO (same for the getNotification)

Example
async function validateQrCodeRequest(request, context) {
if(error){
 return resposeError
}
....
....
return {
    aarQrCodeValue: parsedBody.data.aarQrCodeValue,
    sendHeaders: sendHeadersResult.data,
  };
}
` 

Regarding the BL i think that the parsing of the request is the only BL of this handler.

What do you think?

apps/send-func/src/adapters/functions/__tests__/aar-notifications.test.ts
lucamancosu: Because even if i'm returning HttpResponseInit with my handler, typescript still say that result can be HttpResponse | HttpResponseInit. HttpResponse doesn't have jonBody as prop and it give me an error.

apps/send-func/src/adapters/functions/aar-notifications.ts
lucamancosu: 731d33f38a95d6822337f0628c242b01b3833edb

apps/send-func/src/adapters/functions/aar-notifications.ts
lucamancosu: 731d33f38a95d6822337f0628c242b01b3833edb

apps/send-func/src/adapters/functions/aar-qrcode-check.ts
lucamancosu: 731d33f38a95d6822337f0628c242b01b3833edb

apps/send-func/src/adapters/functions/aar-qrcode-check.ts
lucamancosu: 731d33f38a95d6822337f0628c242b01b3833edb

apps/send-func/src/adapters/functions/aar-qrcode-check.ts
lucamancosu: 731d33f38a95d6822337f0628c242b01b3833edb

apps/send-func/src/adapters/functions/aar-qrcode-check.ts
lucamancosu: 731d33f38a95d6822337f0628c242b01b3833edb

apps/send-func/src/adapters/functions/__tests__/aar-notifications.test.ts
lucamancosu: 1af30a3dae4ef22039b4442c6da65805912e580c

apps/send-func/src/adapters/functions/__tests__/aar-notifications.test.ts
lucamancosu: 1af30a3dae4ef22039b4442c6da65805912e580c

apps/send-func/src/adapters/functions/__tests__/aar-notifications.test.ts
lucamancosu: 1af30a3dae4ef22039b4442c6da65805912e580c

apps/send-func/src/adapters/functions/__tests__/aar-qrcode-check.test.ts
lucamancosu: 1af30a3dae4ef22039b4442c6da65805912e580c

apps/send-func/src/adapters/functions/__tests__/aar-qrcode-check.test.ts
lucamancosu: 1af30a3dae4ef22039b4442c6da65805912e580c

apps/send-func/src/adapters/functions/__tests__/aar-qrcode-check.test.ts
lucamancosu: 1af30a3dae4ef22039b4442c6da65805912e580c

apps/send-func/src/adapters/functions/aar-qrcode-check.ts
lucamancosu: We are using the same structure declared in the openapi. 
We can discuss next week about errors and decide to change it.

apps/send-func/src/adapters/functions/aar-qrcode-check.ts
lucamancosu: I think is related to the same error discussion we should have next week

apps/send-func/src/adapters/functions/aar-notifications.ts
lucamancosu: 731d33f38a95d6822337f0628c242b01b3833edb

apps/send-func/local.settings.json.example
lucamancosu: Can we do it on another task ? 
My proposal is to use, for this release, the production lollipop service

apps/send-func/src/adapters/functions/aar-notifications.ts
Garma00: > Regarding the BL i think that the parsing of the request is the only BL of this handler.

I recognize that this is only a proxy, but I prefer to maintain a clear separation between the request validation step and the core logic.

It would be good to have a generic method that can validate a request payload, perhaps using a Zod schema so we can reuse it.

I'm not comfortable writing the validation directly inside the handler function.

apps/send-func/src/adapters/send/definitions.ts
```diff
@@ -173,3 +173,16 @@ export const sendHeadersSchema = lollipopHeadersSchema.merge(
...
 
 export type SendHeaders = z.TypeOf<typeof sendHeadersSchema>;
+
+export const sendAARClientResponseSchema = z.object({
+  jsonBody: z.union([
+    checkQrMandateResponseSchema,
+    problemJsonSchema,
+    thirdPartyMessageSchema,
+  ]),
+  status: z.number().int(),
```
Garma00, line 185: Can we use a literal here? We can pass any int here and it would be good for the compiler

Garma00 (changes requested):

apps/send-func/src/adapters/send/definitions.ts
lucamancosu: Do you mean a z.union of z.literal ? like 

status: z.union([z.literal(200), z.literal(500)....])

Garma00: Yes, we should only allow status codes the openapi expose

packages/io-messages-common/src/adapters/lollipop/lollipop-middleware.ts
```diff
@@ -134,9 +132,9 @@ export const parseLollipopHeaders = async (
 
 export function createLollipopMiddleware(
   lollipopClient: LollipopClient,
-): Middleware {
-  return async (req: HttpRequest, ctx: InvocationContext) => {
+): Middleware<[LollipopHeaders]> {
+  return async (req: HttpRequest) => {
     const headers = await parseLollipopHeaders(req, lollipopClient);
```
Garma00: I think we should handle the exception here

Garma00 (changes requested):

Garma00 (approved):

packages/io-messages-common/src/adapters/lollipop/lollipop-middleware.ts
acuiuli: it is already handled in the `handlerWithMiddleware` funct

PR #432, opened 2025-09-30
Update SEND AAR openapi

lucamancosu:


Vangaorth (changes requested):
The same schema extraction that has been done for TaxId should also be applied to the `iun` property of `AARQRCodeCheckResponse`

Garma00 (approved):

Vangaorth (approved):
LGTM

PR #433, opened 2025-09-30
Create / delete installation on both WEU and ITN notification hub

robbcocco:
Closes: IOCOM-2607


Garma00:
@robbcocco please can you add some context in the PR description? Why do we need those changes?

apps/pushnotif-func/src/utils/notification.ts
```diff
@@ -133,47 +133,120 @@ export const notify = (
...
 
+type MaybeErrorResponse = Error | NotificationHubsResponse;
+
 export const createOrUpdateInstallation = (
   notificationHubClient: NotificationHubsClient,
+  legacyNotificationHubClient: NotificationHubsClient,
   installationId: NonEmptyString,
   platform: Platform,
   pushChannel: string,
   tags: readonly string[],
```
Garma00, line 142: I think that the new logic of this function (and te deleteInstallation as well) is a bit too complex to handle a temporary problem.

May I suggest you to just accept a TelemetryClient as input parameter and use it directly when the second createOrUpdate fails?

This way: 

1. You don't need to create a new type `MaybeErrorResponse`
2. The Return type of the function does not change as well as the logic the caller needs to implement
3. Here you just need to track an exception in case the second createOrUpdateFails

Take a look at this implementation
```ts
export const createOrUpdateInstallation = (
  notificationHubClient: NotificationHubsClient,
  legacyNotificationHubClient: NotificationHubsClient,
  installationId: NonEmptyString,
  platform: Platform,
  pushChannel: string,
  tags: readonly string[],
  telemetryClient: TelemetryClient,
): TE.TaskEither<Error, NotificationHubsResponse> =>
  pipe(
    TE.tryCatch(
      () =>
        legacyNotificationHubClient.createOrUpdateInstallation(
          // create the installation in the first NH
       ),
       // we return an error in case the first fails
      (errs) =>
        new Error(
          `Error while creating or updating installation on NotificationHub [${errs}]`,
        ),
    ),
    // Here we use chainFirst because, no matter the result of the second creation, 
    // we still want to return a success because the first call was a success, 
    // we just want to track an exception in case of an error during the second creation
    // (please note that we want the sampling disabled here)
    TE.chainFirst(() =>
      TE.tryCatch(
        () =>
          notificationHubClient.createOrUpdateInstallation({
            // create or update on the second NH
          }),
        (err) => {
          telemetryClient.trackException({
            exception: new Error(
              `Error while creating or updating installation on new NotificationHub [${installationId}] [${err}]`,
            ),
            properties: {
              isSuccess: "false",
              name: "api.messages.notification.createOrUpdateInstallation.newNh.error",
            },
            tagOverrides: { samplingEnabled: "false" },
          });
          return new Error(
            `Error while creating or updating installation on NotificationHub [${err}]`,
          );
        },
      ),
    ),
  );
```

apps/pushnotif-func/src/utils/notification.ts
```diff
@@ -133,47 +133,120 @@ export const notify = (
...
+      );
+    }),
   );
 
 export const deleteInstallation = (
-  notificationHubService: NotificationHubsClient,
+  notificationHubClient: NotificationHubsClient,
+  legacyNotificationHubClient: NotificationHubsClient,
   installationId: NonEmptyString,
-): TaskEither<Error, NotificationHubsResponse> =>
```
Garma00, line 172: If you agree with the commend above, I would do the same here

Garma00 (changes requested):

infra/resources/_modules/web_apps/pushnotif-func.tf
```diff
@@ -99,23 +139,47 @@ locals {
...
+      NH3_NAME            = data.azurerm_notification_hub.common_partition_itn[2].name
       NH4_PARTITION_REGEX = "^[c-f]"
-      NH4_NAME            = data.azurerm_notification_hub.common_partition[3].name
+      NH4_NAME            = data.azurerm_notification_hub.common_partition_itn[3].name
+
+      AZURE_NH_ENDPOINT = data.azurerm_key_vault_secret.azure_nh_endpoint_itn.value
+      NH1_ENDPOINT      = data.azurerm_key_vault_secret.azure_nh_partition1_endpoint_itn.value
+      NH2_ENDPOINT      = data.azurerm_key_vault_secret.azure_nh_partition2_endpoint_itn.value
+      NH3_ENDPOINT      = data.azurerm_key_vault_secret.azure_nh_partition3_endpoint_itn.value
+      NH4_ENDPOINT      = data.azurerm_key_vault_secret.azure_nh_partition4_endpoint_itn.value
```
Garma00: We don't need to use a data to get those vaules, we can get them directly from the module.
I know that for the old ones we handled it like that but I prefer a different approach.
May I suggest you to take those values as input variables?
```suggestion
      NH1_ENDPOINT      = var.nh_partition_1.endpoint
      NH2_ENDPOINT      = var.nh_partition_2.endpoint
      NH3_ENDPOINT      = var.nh_partition_3.endpoint
      NH4_ENDPOINT      = var.nh_partition_4.endpoint
```

Then we can pass those values in `prod/web_apps.tf` with something like this
```tf
nh_partition_1 = module.notification_hubs_itn.partition_1
```
Probably we need to add an `output.tf` in `_modules/notification_hubs_itn`

infra/resources/_modules/web_apps/pushnotif-func.tf
```diff
@@ -99,23 +139,47 @@ locals {
...
-      NH2_NAME            = data.azurerm_notification_hub.common_partition[1].name
+      NH2_NAME            = data.azurerm_notification_hub.common_partition_itn[1].name
       NH3_PARTITION_REGEX = "^[8-b]"
-      NH3_NAME            = data.azurerm_notification_hub.common_partition[2].name
+      NH3_NAME            = data.azurerm_notification_hub.common_partition_itn[2].name
       NH4_PARTITION_REGEX = "^[c-f]"
-      NH4_NAME            = data.azurerm_notification_hub.common_partition[3].name
+      NH4_NAME            = data.azurerm_notification_hub.common_partition_itn[3].name
+
+      AZURE_NH_ENDPOINT = data.azurerm_key_vault_secret.azure_nh_endpoint_itn.value
```
Garma00:
```suggestion
```

I think this is not necessary, we only have 4 partitions in itn now

Garma00 (changes requested):

infra/resources/_modules/web_apps/pushnotif-func.tf
```diff
@@ -99,23 +139,47 @@ locals {
...
+      LEGACY_NH3_PARTITION_REGEX = "^[8-b]"
+      LEGACY_NH3_NAME            = data.azurerm_notification_hub.common_partition[2].name
+      LEGACY_NH4_PARTITION_REGEX = "^[c-f]"
+      LEGACY_NH4_NAME            = data.azurerm_notification_hub.common_partition[3].name
+
+      LEGACY_AZURE_NH_ENDPOINT = data.azurerm_key_vault_secret.azure_nh_endpoint.value
+      LEGACY_NH1_ENDPOINT      = data.azurerm_key_vault_secret.azure_nh_partition1_endpoint.value
+      LEGACY_NH2_ENDPOINT      = data.azurerm_key_vault_secret.azure_nh_partition2_endpoint.value
+      LEGACY_NH3_ENDPOINT      = data.azurerm_key_vault_secret.azure_nh_partition3_endpoint.value
+      LEGACY_NH4_ENDPOINT      = data.azurerm_key_vault_secret.azure_nh_partition4_endpoint.value
```
Garma00: ditto

Garma00 (changes requested):

infra/resources/_modules/web_apps/pushnotif-func.tf
robbcocco: i would leave the legacy config as is since it's been working fine and it will be removed after the migration is completed

infra/resources/_modules/notification_hubs_itn/outputs.tf
```diff
@@ -0,0 +1,29 @@
+output "nh_itn_partition_1" {
+  value = {
+    id       = azurerm_notification_hub.partition_1.id
+    name     = azurerm_notification_hub.partition_1.name
+    endpoint = azurerm_notification_hub_authorization_rule.partition_1_default_full.primary_connection_string
+    # endpoint = "https://${azurerm_notification_hub.partition_1.namespace_name}.servicebus.windows.net/${azurerm_notification_hub.partition_1.name}/"
```
Garma00:
```suggestion
```

infra/resources/_modules/web_apps/pushnotif-func.tf
Garma00: You are right, nice point

Garma00 (changes requested):

Garma00 (approved):
Well done @robbcocco! I just tested locally the new logic and it works as expected, we are only duplicating the create and the delete operations on the new notification hubs, if I try to send a notification it is only sent one time 👌🏽

PR #437, opened 2025-10-03
new getAttachment api

acuiuli:
Add the new api to get the attachment metadata for a send AAR

Closes: [IOCOM-2580](https://pagopa.atlassian.net/browse/IOCOM-2580)

[IOCOM-2580]: https://pagopa.atlassian.net/browse/IOCOM-2580?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/send-func/src/adapters/functions/aar-attachments.ts
+++ b/apps/send-func/src/adapters/functions/aar-attachments.ts
@@ -68,10 +68,7 @@ export const getAttachment =
       );
       return { jsonBody: response, status: 200 };
     } catch (err) {
-      if (
-        err instanceof NotificationClientError &&
-        err.name === "NotificationClientError"
-      ) {
+      if (err instanceof NotificationClientError) {
         context.error("Notification client error:", err.message);
 
         const problemJson = problemJsonSchema.parse(err.body);
--- a/apps/send-func/src/adapters/functions/aar-attachments.ts
+++ b/apps/send-func/src/adapters/functions/aar-attachments.ts
@@ -1,0 +1,163 @@
+import {
+  attachmentNameSchema,
+  idxSchema,
+  iunSchema,
+  mandateIdSchema,
+} from "@/domain/notification.js";
+import {
+  AttachmentParams,
+  GetAttachmentUseCase,
+  attachmentParamsSchema,
+} from "@/domain/use-cases/get-attachment.js";
+import { HttpRequest, InvocationContext } from "@azure/functions";
+import { LollipopHeaders } from "io-messages-common/adapters/lollipop/definitions/lollipop-headers";
+import { ExtentedHttpHandler } from "io-messages-common/adapters/middleware";
+
+import {
+  AarGetAttachmentResponse,
+  problemJsonSchema,
+} from "../send/definitions.js";
+import { NotificationClientError } from "../send/notification.js";
+import { malformedBodyResponse } from "./commons/response.js";
+
+export const getAttachment =
+  (
+    getAttachmentUseCase: GetAttachmentUseCase,
+  ): ExtentedHttpHandler<LollipopHeaders> =>
+  async (
+    request: HttpRequest,
+    context: InvocationContext,
+    lollipopHeaders: LollipopHeaders,
+  ): Promise<AarGetAttachmentResponse> => {
+    const isTest = request.query.get("isTest") === "true";
+
+    const sendHeaders = {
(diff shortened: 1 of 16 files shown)


apps/send-func/src/func.ts
```diff
@@ -54,11 +56,21 @@ const main = async (config: Config): Promise<void> => {
...
   });
+
+  app.http("getNotificationAttachment", {
+    authLevel: "anonymous",
+    handler: handlerWithMiddleware(
+      lollipopMiddleware,
+      getAttachment(getAttachmentUseCase),
+    ),
+    methods: ["GET"],
+    route: "/aar/attachments/{attachmentUrl}",
```
Garma00:
```suggestion
    route: "aar/attachments/{attachmentUrl}",
```

apps/send-func/src/adapters/functions/aar-attachments.ts
```diff
@@ -0,0 +1,163 @@
...
+  urlEncodedBase64Url: string,
+):
+  | { data: AttachmentParams; success: true }
+  | { error: Error; success: false } {
+  try {
+    const base64Url = decodeURIComponent(urlEncodedBase64Url);
+    const base64UrlBuffer = Buffer.from(base64Url, "base64");
+    const attachmentUrl = base64UrlBuffer.toString("utf8");
+
+    const [path, queryString] = attachmentUrl.split("?");
```
Garma00, line 102: Are we sure that the queryString will always be the 2nd output of this split?

apps/send-func/src/adapters/functions/aar-attachments.ts
```diff
@@ -0,0 +1,163 @@
...
+        attachmentParams.data,
+        sendHeaders,
+        isTest,
+        parsedMandateId.data,
+      );
+      return { jsonBody: response, status: 200 };
+    } catch (err) {
+      if (
+        err instanceof NotificationClientError &&
+        err.name === "NotificationClientError"
```
Garma00:
```suggestion
        err instanceof NotificationClientError
```

Is the double check needed?

apps/send-func/src/adapters/functions/aar-attachments.ts
```diff
@@ -0,0 +1,163 @@
...
+      ) {
+        context.error("Notification client error:", err.message);
```
Garma00, line 72: Maybe tracking an event/exception is better here but I think we can make an in-depth analysis later

Garma00 (changes requested):

apps/send-func/src/adapters/functions/aar-attachments.ts
lucamancosu: Can we use the builtin URL class here ? 

something like:

const parsedUrl = new URL(myURL)

If i remember well parsedURL is an object containing all the properties you need

lucamancosu (approved):
Good job!!!

apps/send-func/src/adapters/functions/aar-attachments.ts
acuiuli: We cannot use the URL built in class because that is not really an URL but it's just a path (i.e. `/delivery/notifications/received/{iun}/attachments/payment/{attachmentName}` with an optional query param and not a full URL (i.e. `https://.....`) like it is required to create an URL.
In regards to the order, yes we are sure that it is the second output of the split

apps/send-func/src/adapters/functions/aar-attachments.ts
acuiuli: no, [here](https://github.com/pagopa/io-messages/pull/437#:~:text=properties%20you%20need-,fix%20catch%20error,-00ec83d)

apps/send-func/src/adapters/functions/aar-attachments.ts
acuiuli: we have a task for the error/monitoring part and we will do it later after we do an in-depth analysis. I cose this

apps/send-func/src/adapters/functions/aar-attachments.ts
Garma00: > yes we are sure that it is the second output of the split

Thanks @acuiuli, can we add a comment with this rational?

Garma00 (approved):

PR #441, opened 2025-10-07
Update send openapi and apim policy

lucamancosu:


apps/send-func/openapi/aar-notification.yaml
Garma00, line 1: I suggest to add the `/aar` to the base path instead of repeating ing to every path if this is the yaml specific to the aar

Garma00 (changes requested):

apps/send-func/openapi/aar-notification.yaml
lucamancosu: 0204618bba5c936eff4677627087e5269c3f33da

Garma00 (approved):

PR #452, opened 2025-10-09
Create new P3 redis

Garma00:
Directly upgrading the redis would cause downtime during the apply, so we first create a new redis instance.

Closes IOCOM-2646

### DIFF
--- a/infra/resources/prod/redis.tf
+++ b/infra/resources/prod/redis.tf
@@ -59,3 +59,66 @@ resource "azurerm_private_endpoint" "redis_cache_com" {
 
   tags = local.tags
 }
+
+# this resource name will be updated once we will remove the old one
+resource "azurerm_redis_cache" "com_new" {
+
+  name = provider::dx::resource_name({
+    prefix          = local.prefix
+    name            = "com",
+    resource_type   = "redis_cache",
+    environment     = local.env_short,
+    location        = local.location
+    instance_number = 2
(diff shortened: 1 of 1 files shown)


infra/resources/prod/redis.tf
```diff
@@ -12,9 +12,9 @@ resource "azurerm_redis_cache" "com" {
   resource_group_name = azurerm_resource_group.itn_com.name
   location            = azurerm_resource_group.itn_com.location
 
-  capacity            = 2
-  family              = "C"
-  sku_name            = "Standard"
+  capacity            = 3
+  family              = "P"
+  sku_name            = "Premium"
```
lucacavallaro: this operation can cause downtime and slow response from redis, it is better to create a new resource

Garma00: Thanks for the suggestion, now we are just creating the new redis [ab6a9ca](https://github.com/pagopa/io-messages/pull/452/commits/ab6a9caaf0058ea153d69c42bf90345a1f9b7953)

lucamancosu (approved):

Garma00:
> We need to update the infra/resources/prod/web_apps.tf file with the new redis ? Probably is enough but is better to do a check on all the services that use the com redis.

Yess but it is better to create the new redis first, I would make those changes in a separate PR @lucamancosu

PR #453, opened 2025-10-09
Use new redis into web_apps module

Garma00:
- Use the new `io-p-itn-com-redis-02` for the `web_apps`
- Remove useless `azurerm_redis_cache_access_policy_assignment` for the `etl-func`

Closes IOCOM-2646


Garma00:
It seems we have a `azurerm_redis_cache_access_policy_assignment` for the `etl-func` only, I think we can remove it cause we are using the access key to authenticate (in fact it works perfectly for the others functions) cc @acuiuli @lucamancosu

lucamancosu (approved):

PR #454, opened 2025-10-09
Add cli for exporting / importing registrations and installations between two notification hubs

robbcocco:
Closes IOCOM-2608


packages/io-nh-utils/.gitignore
Garma00, line 1: We should use the `.gitignore` in the root of the monorepo

packages/io-nh-utils/src/import.ts
```diff
@@ -0,0 +1,90 @@
...
+  const fromClient = new NotificationHubsClient(
+    fromNotificationHub.connectionString,
+    fromNotificationHub.hubName,
+  );
+  const toClient = new NotificationHubsClient(
+    toNotificationHub.connectionString,
+    toNotificationHub.hubName,
+  );
+
+  const installationIds = [...new Set(rows.map((r) => r.installationId))];
```
Garma00: Can you add a comment that explain the need of using a set here to give the reader some context?

packages/io-nh-utils/src/import.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  for await (const installationId of installationIds) {
+    await migrateInstallation(fromClient, toClient, installationId);
+    if (installationIds.indexOf(installationId) % 100 === 0) {
+      //eslint-disable-next-line no-console
+      console.log(
+        `${new Date(Date.now()).toLocaleString("it-IT")} - Imported ${installationIds.indexOf(installationId)} installations...`,
+      );
+    }
```
Garma00: @robbcocco What if `migrateInstallation` throw here? Here the program would crash, I think we want to handle the exception by printing an error and skip the iteration, what's your take?

packages/io-nh-utils/src/export.ts
```diff
@@ -0,0 +1,97 @@
...
+      continuationToken = result.continuationToken;
+    } else {
+      const result = await run({
+        connectionString,
+        hubName,
+      });
+      rows = result.rows;
+    }
+
+    await saveCsv(rows, outputPath());
```
Garma00: I would refactor this in order to append the rows to the `csv` at every iteration and print the continuation token, this way if the program exit before the export ended, we don't lose the already exported installations and we can resume the export operation using the last continuation token printed in the stdout.

Garma00 (changes requested):

packages/io-nh-utils/.gitignore
robbcocco: since we're gitignoring a directory specific only to this project i think it's cleaner to use a local gitignore, but if we only want to use the gitignore from the workspace root we could either use a path with wildcards `**/data/*.csv` or the path `packages/io-nh-utils/data`. former seems too generic and could maybe impact future folders? i think the latter is best

packages/io-nh-utils/src/import.ts
robbcocco: resolved in ae7f2be6c75f4af57daa6e1d9b024a5ea74053c1

packages/io-nh-utils/src/export.ts
robbcocco: resolved in ae7f2be6c75f4af57daa6e1d9b024a5ea74053c1

packages/io-nh-utils/src/import.ts
robbcocco: resolved in ae7f2be6c75f4af57daa6e1d9b024a5ea74053c1

PR #457, opened 2025-10-09
Error management AAR SEND API

acuiuli:
- Update zod to version 4 to better manage definitions and ZodError
- fix type parsing in notification client


Garma00 (approved):
LGTM apart of the format script failing

PR #461, opened 2025-10-13
Enable Ingest Message Status

robbcocco:
Update start from time to 2023/01/01

### DIFF
--- a/infra/resources/_modules/web_apps/etl_func.tf
+++ b/infra/resources/_modules/web_apps/etl_func.tf
@@ -47,10 +47,12 @@ module "etl_func" {
 
   health_check_path = "/api/health"
 
-  app_settings      = local.etl_func.app_settings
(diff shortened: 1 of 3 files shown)


infra/resources/_modules/web_apps/etl_func.tf
```diff
@@ -2,7 +2,6 @@ locals {
   etl_func = {
     app_settings = {
       NODE_ENV                                       = "production",
-      "AzureWebJobs.IngestMessageStatus.Disabled"    = "1",
```
lucamancosu, line 5: We must work with sticky app settings, if i remember well, because we must turn on the message status ingestion in the production slot only. Take a look here infra/resources/_modules/web_apps/pushnotif-func.tf for example

Garma00 (approved):

PR #465, opened 2025-10-13
Make send-func testable locally

Garma00:
- Add a new send mock using mockoon
- Add a new bruno collection with all the api request needed to test the send-func locally


Garma00:
@lucamancosu here we can use a bruno script to populate the lollipop headers, then we can remove the useless bruno secrets, at the moment we need them to test this function locally.

lucamancosu (approved):

PR #477, opened 2025-10-20
Update CODEOWNERS to scope devex team to bootstrapper folder

lucacavallaro:


lucacavallaro:
@pagopa/io-communication-backend I noticed that the `CODEOWNER` file explicitly mentions `@lucamancosu` - it is intentional or can be replaced with `@pagopa/io-communication-backend-ext` group? If so, please do it within a subsequent PR.

Garma00 (approved):

PR #529, opened 2025-12-03
Switch to IO COM blob storage in services func

robbcocco:
Closes IOCOM-2821


Garma00 (changes requested):
Well done! We have some dependencies tho, we can't merge it yet

PR #532, opened 2025-12-04
Add init azurite configuration

robbcocco:
Added a script to initialize containers, queues and tables in azurite.


azurite-containers.js
```diff
@@ -0,0 +1,49 @@
...
+  "AccountName=devstoreaccount1;" +
+  "AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;" +
+  "BlobEndpoint=http://azurite:10000/devstoreaccount1;" +
+  "QueueEndpoint=http://azurite:10001/devstoreaccount1;";
+
+const blobServiceClient =
+  BlobServiceClient.fromConnectionString(connectionString);
+
+const queueServiceClient =
+  QueueServiceClient.fromConnectionString(connectionString);
```
Garma00: Please also add the tables we need (for `etlf-func` in example).

azurite-containers.js
```diff
@@ -0,0 +1,49 @@
...
+    await containerClient.createIfNotExists();
+  }
+}
+
+async function ensureQueues() {
+  for (const name of queues) {
+    const queueClient = queueServiceClient.getQueueClient(name);
+    await queueClient.createIfNotExists();
+  }
+}
```
Garma00: I think we can make it a bit faster with `Promise.all` instead of waiting for the previous creation to finish, please make this change also for the `ensureContainers` and for the`ensureTables`

azurite-containers.js
```diff
@@ -0,0 +1,49 @@
...
+
+(async () => {
+  await ensureContainers();
+  await ensureQueues();
+})();
```
Garma00: Maybe it can be useful to add some logs

azurite-containers.js
```diff
@@ -0,0 +1,49 @@
+const {
```
Garma00: Can we add a comment with a little description explaining why this file is needed and what problem does it solves?

Garma00 (changes requested):
Also I would rename the `azurite-containers.js` file using something more meaningful such as `setup-azurite-container.js` or something else

Garma00 (approved):

PR #534, opened 2025-12-09
Add SEND AAR error response data to send-func telemetry

lucamancosu:
Closes: IOCOM-2812


Garma00 (changes requested):
I don't like logging unknown response objects because we risk logging sensitive data. I suggest closing this one and opening another PR where we only track events with known flags (such as traceId).

Garma00 (approved):
LGTM but I think it is better to release a patch instead of minor.

PR #549, opened 2026-01-13
Improve services-func local testability

Garma00:
- [Create apim mock with mockoon](https://github.com/pagopa/io-messages/commit/52113817577c79f7895bf1fc203f7e9420f118bd)
- [Remove redundant bruno variables](https://github.com/pagopa/io-messages/commit/620d8bf2ee87bff07bac0409aeaaa345dfddb57f)

NB: For the moment this mock only handles the happy path of the
getRCConfiguration call, this can be extended in future.

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -73,6 +73,14 @@ services:
     volumes:
       - "./mockoon/session-manager.json:/data:readonly"
 
+  mock-server-apim:
+    image: mockoon/cli:latest@sha256:80c4acd9517840b6574784d420de2dc97a8b750d21b6a4b573d042d21e4833ba
+    command: ["--data", "data", "--port", "3000"]
+    ports:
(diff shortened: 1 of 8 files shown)


robbcocco (approved):

compose.yml
```diff
@@ -73,6 +73,14 @@ services:
     volumes:
       - "./mockoon/session-manager.json:/data:readonly"
 
+  mock-server-apim:
```
lucamancosu, line 76: LGTM. If you want you can just change it in services-apim just to be clear about which APIM we are mocking.

lucamancosu (approved):

compose.yml
Garma00: I prefer to leave it general so we can try to reuse this container for other micro-services if we need 😃

PR #560, opened 2026-01-20
Add send-func info API

lucamancosu:
Close: IOCOM-2893

The API returns the name and the version of the function


Garma00 (approved):
LGTM!

PR #571, opened 2026-02-05
Add SEND lollipop integration check API to APIM

lucamancosu:
Closes: IOCOM-2920

### DIFF
--- a/infra/resources/_modules/apim/send.tf
+++ b/infra/resources/_modules/apim/send.tf
@@ -110,7 +110,7 @@ resource "azurerm_api_management_api" "send_lollipop_integration_check_api_v1" {
 
 
 resource "azurerm_api_management_api_policy" "send_lollipop_integration_check_api_v1_policy" {
-  api_name = azurerm_api_management_api.send_api_v1.name
+  api_name = azurerm_api_management_api.send_lollipop_integration_check_api_v1.name
 
   api_management_name = data.azurerm_api_management.apim_itn_platform_api.name
   resource_group_name = data.azurerm_api_management.apim_itn_platform_api.resource_group_name
--- a/infra/resources/_modules/apim/send.tf
+++ b/infra/resources/_modules/apim/send.tf
@@ -100,7 +100,6 @@ resource "azurerm_api_management_api" "send_lollipop_integration_check_api_v1" {
   protocols = ["https"]
 
   subscription_required = false
-  service_url           = null
 
   import {
     content_format = "openapi-link"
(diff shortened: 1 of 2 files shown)


infra/resources/_modules/apim/send.tf
```diff
@@ -80,3 +80,48 @@ resource "azurerm_api_management_product_api" "pn_api_v1_product_api" {
...
+}
+
+
+resource "azurerm_api_management_api_policy" "send_lollipop_integration_check_api_v1_policy" {
+  api_name = azurerm_api_management_api.send_lollipop_integration_check_api_v1.name
+
+  api_management_name = data.azurerm_api_management.apim_itn_platform_api.name
+  resource_group_name = data.azurerm_api_management.apim_itn_platform_api.resource_group_name
+
+  xml_link = "https://raw.githubusercontent.com/pagopa/io-messages/060caa4dc8edcc2f1e9200750b917dd88254dfc8/infra/resources/_modules/apim/api/send/policy.xml" #-- Da aggiornare con policy specifica
```
giamma1295: Is the definitive policy the one included in [this PR](https://github.com/pagopa/io-messages/pull/572)?

Why do we do two PRs instead of just one with both changes, linking the policy in a similar way to:

`xml_content = file("${path.module}/path/to/lollipop-integration-check-policy.xml/_pn_communications_policy.xml")`?

lucamancosu: Two reasons:
1. using xml_link we can pointing to a specific file in the history avoiding unwanted changes 
2. using xml_content, instead of xml_link, gave drifts problems for indentations reasons executing terraform plans

infra/resources/_modules/apim/send.tf
```diff
@@ -80,3 +80,48 @@ resource "azurerm_api_management_product_api" "pn_api_v1_product_api" {
...
+
+  contact {
+    name = "pagoPA - Touchpoints team"
+  }
+
+  path      = "api/com/v1/send/lollipop-check"
+  protocols = ["https"]
+
+  subscription_required = false
+  service_url           = null
```
Garma00: I think we can remove it
```suggestion
```

infra/resources/_modules/apim/send.tf
giamma1295: 1. Since the policy is specific to this openAPI, in what scenario do we plan to update policies while keeping the previous version active to avoid unwanted changes?
2. do you have an example of drift problems for indentations on terraform plan? According to this PR's tf plan there aren't drift related to these https://github.com/search?q=repo%3Apagopa%2Fio-messages%20xml_content%20%3D%20file&type=code

lucamancosu: 1. right now i dont have an example but is safer anyway to point to a specific commit 
2. we had this kind of problem in the past but retrieve an old and exact PR is not so simple

Garma00 (approved):

PR #587, opened 2026-02-17
Update etl-func node version

lucamancosu:
Upgrade etl-func node version runtime from 20 to 22 LTS.

Successfully tested the etl-func triggers locally under the node 22 LTS  runtime.

Closes: IOCOM-2978


giamma1295 (approved):
LGTM

PR #593, opened 2026-02-23
Add bruno configuration for ops-func

Garma00:


bruno/ops-func/Health.bru
```diff
@@ -0,0 +1,11 @@
+meta {
```
lucamancosu, line 1: This is an empty request

lucamancosu (approved):

PR #594, opened 2026-02-24
rc-func to programming model v4

lucamancosu:
Tested all the http triggers and the cosmosdb trigger locally.

Closes: IOCOM-2959


apps/rc-func/package.json
```diff
@@ -43,10 +44,10 @@
     "@azure/identity": "^4.5.0",
     "@azure/storage-queue": "^12.4.0",
     "@pagopa/express-azure-functions": "^2.1.1",
-    "@pagopa/io-functions-commons": "^29.3.0",
+    "@pagopa/io-functions-commons": "^30.0.0",
     "@pagopa/ts-commons": "^12.5.0",
```
giamma1295: ts-commons should be updated to v13 eg `"^13.1.2"` to match `io-functions-commons ^30.x.x` peerDeps requirement!

See https://github.com/pagopa/io-functions-commons/blob/faf6896433fd23ad60d5fa59f5c432692fb13b79/package.json#L86

lucamancosu: 3fb39fb4597c8925e03b32f2e8c597af05ab8b16

giamma1295 (approved):
LGTM

PR #600, opened 2026-02-25
Remove unused session fragment 

lucamancosu:
The fragment removed with this PR is already replaced by the **ioapp-authenticated** fragment in production. 
The removal of this fragment does not effect any API on the APIM

Closes: IOCOM-3003


giamma1295 (approved):
LGTM

PR #611, opened 2026-02-27
Add AGENTS.md

giamma1295:
Add Agents instruction into monorepo

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -102,6 +102,8 @@ yarn workspace <workspace-name> add <package>
 yarn workspace <workspace-name> add -D <package>
 ```
 
+**Linting operations** on `apps` or `package` edit always perform a lint on modified workspaces.
+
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,107 @@
...
+# Lint a single workspace
+yarn workspace <workspace-name> lint
+
+```
+
+Turbo caches outputs; use `--force` to bust the cache when needed.
+
+## Architecture
+
+This is a **Yarn workspaces + Turborepo monorepo** hosting multiple Azure Functions apps and a shared package.
```
lucamancosu: can be useful tell that we are using Yarn PnP ?

lucamancosu (approved):

AGENTS.md
giamma1295: added!

PR #635, opened 2026-03-11
SEND AAR remove x-pagopa-pn-io-src hardcoded value

lucamancosu:
All SEND AAR APIs now forward the x-pagopa-pn-io-src header to the SEND backend only when it is provided by the mobile app.


Garma00 (approved):
LGTM!

PR #644, opened 2026-03-12
Update error types for the notification repository

Garma00:
- Add `ErrorTooManyRequests`
- Make updateInstallation return `ErrorTooManyRequests`
- Make updateInstallation return `ErrorInternal`
- Remove `ZodError` as return type of updateInstallation (never happens)


giamma1295 (approved):
LGTM

PR #648, opened 2026-03-12
pushnotif: Obtain database name for installation summary from config

robbcocco:


Garma00 (changes requested):
We also need to add the env variable in the infrastructure

Garma00 (approved):

Garma00 (approved):

PR #677, opened 2026-03-26
Separate Health and Info endpoint in Pushnotif func

robbcocco:


apps/pushnotif-func/src/adapters/functions/health.ts
```diff
@@ -0,0 +1,34 @@
...
+import { HttpHandler } from "@azure/functions";
+
+import { ErrorInternal } from "../../domain/error";
+
+export const getHealthHandler =
+  (healthChecks: HealthCheck[]): HttpHandler =>
+  async () => {
+    try {
+      const healthChecksResults = await Promise.all(
+        healthChecks.map((hc) => hc()),
```
Garma00:
```suggestion
        healthChecks.map(hc),
```

Garma00 (changes requested):

apps/pushnotif-func/src/adapters/functions/health.ts
robbcocco: resolved in cbd4c3e3fa7a8540c649b0fd7aa5ec80eb4322aa (map was not needed at all)

Garma00 (approved):

PR #679, opened 2026-03-26
Add create massive notification job function

robbcocco:
Closes 3064


apps/pushnotif-func/src/adapters/cosmos/massive-jobs.ts
```diff
@@ -0,0 +1,38 @@
...
+    } catch (err) {
+      if (err instanceof ErrorResponse) {
+        switch (err.code) {
+          case 404:
+            return new ErrorNotFound("Massive job not found", err);
+          default:
+            return new ErrorInternal("Failed to update massive job", err);
+        }
+      }
+      return new ErrorInternal("Failed to update massive job", err);
```
Garma00, line 32:
```suggestion
      }
      
      return new ErrorInternal("Failed to update massive job", err);
```

robbcocco: Nice catch!

apps/pushnotif-func/src/adapters/functions/create-massive-notification-job.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    if (parsedBody instanceof ErrorValidation) {
+      return {
+        body: JSON.stringify({
+          error: parsedBody.message,
+          issues: parsedBody.issues,
+        }),
+        headers: { "Content-Type": "application/json" },
+        status: 400,
+      };
```
Garma00: Maybe we can create a utility function that accept a body and return an object with headers and status 400 @robbcocco

apps/pushnotif-func/src/adapters/functions/create-massive-notification-job.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    const result =
+      await createMassiveNotificationJobUseCase.execute(parsedBody);
+
+    if (result instanceof ErrorInternal) {
+      return {
+        body: JSON.stringify({ error: result.message }),
+        headers: { "Content-Type": "application/json" },
+        status: 500,
+      };
```
Garma00: DItto

apps/pushnotif-func/src/adapters/functions/create-massive-notification-job.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    return {
+      body: JSON.stringify({ id: result }),
+      headers: { "Content-Type": "application/json" },
+      status: 201,
```
Garma00: Ditto

apps/pushnotif-func/src/domain/use-cases/create-massive-notification-job.ts
```diff
@@ -0,0 +1,27 @@
...
+    massiveJob: CreateMassiveJobPayload,
+  ): Promise<ErrorInternal | ErrorValidation | string> {
+    const job = {
+      body: massiveJob.body,
+      executionTimeInHours: massiveJob.executionTimeInHours,
+      id: ulid(),
+      startTimeTimestamp: massiveJob.startTimeTimestamp,
+      status: "CREATED" as MassiveJobStatus,
+      title: massiveJob.title,
+    };
```
Garma00, line 20: What about this approach?

```suggestion
    const job = {
      ...
      id: ulid()
    };
```

apps/pushnotif-func/src/domain/http-request.ts
```diff
@@ -0,0 +1,30 @@
+// TODO: move this file to io-messages-common
+
+import { HttpRequest } from "@azure/functions";
+import { ZodObject } from "zod/v4/classic/external.cjs";
+
+import { ErrorValidation } from "./error";
+
+//eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const parseRequestBody = async <T extends ZodObject<any>>(
```
Garma00:
```suggestion
export const parseRequestBody = async <T extends z.ZodTypeAny>>(
```

apps/pushnotif-func/src/domain/massive-jobs.ts
```diff
@@ -0,0 +1,43 @@
...
+  "CREATED",
+  "PROCESSING",
+  "COMPLETED",
+  "STOPPED",
+  "FAILED",
+]);
+export type MassiveJobStatus = z.infer<typeof MassiveJobStatusEnum>;
+
+export const MassiveJobSchema = z.object({
+  body: z.string().min(1),
```
Garma00: We may want to add a `max`, since this is the message that will appears in the push we don't want it to be so long. @Vangaorth What is your take?

apps/pushnotif-func/src/domain/massive-jobs.ts
```diff
@@ -0,0 +1,43 @@
...
+  executionTimeInHours: z.number().int().min(2).max(12).default(2),
+  id: z.uuid(),
+  startTimeTimestamp: z
+    .number()
+    .int()
+    .positive()
+    .default(() => Math.floor((Date.now() + 60 * 60 * 1000) / 1000)), // default to 1 hour from now
+  status: MassiveJobStatusEnum,
+  title: z.string().min(1),
```
Garma00: Ditto cc @Vangaorth

apps/pushnotif-func/src/main.ts
```diff
@@ -287,6 +290,22 @@ const main = (config: Config) => {
...
+
+  const massiveJobsRepository = new CosmosMassiveJobsAdapter(
+    pushCosmosDb,
+    config.massiveJobsContainerName,
+  );
+  const createMassiveNotificationJobUseCase =
+    new CreateMassiveNotificationJobUseCase(massiveJobsRepository);
+
+  app.http("CreateMassiveNotificationJob", {
+    authLevel: "admin",
```
Garma00, line 304: Is this auth level ok for our use case? cc @giamma1295

Garma00 (changes requested):

apps/pushnotif-func/src/domain/massive-jobs.ts
Vangaorth: I think we should only concern ourselves with the technical limitations, meaning our only check should be ensuring that the entire notification payload does not exceed the 4KB limit (maybe set this as an env variable).

As for the title and body display, it depends so heavily on OS behavior, device size, accessibility settings, and so on, that we shouldn't try to guess text limits. We should leave that decision up to whoever designs the notification's content.

apps/pushnotif-func/src/domain/massive-jobs.ts
Vangaorth: See my reply above

apps/pushnotif-func/src/domain/massive-jobs.ts
Garma00: Ok, thanks!

apps/pushnotif-func/src/adapters/functions/create-massive-notification-job.ts
robbcocco: Resolved in 6b8f5e07dd094ca437cc43a2ed3dd7519f8b1988

apps/pushnotif-func/src/domain/use-cases/create-massive-notification-job.ts
robbcocco: Restored in 6b8f5e07dd094ca437cc43a2ed3dd7519f8b1988

apps/pushnotif-func/src/domain/http-request.ts
robbcocco: Applied in 6b8f5e07dd094ca437cc43a2ed3dd7519f8b1988

Garma00 (approved):

PR #705, opened 2026-04-09
Add StartMassiveJob function in pushnotif

robbcocco:
Closes IOCOM-3065


apps/pushnotif-func/docs/puml/create-massive-job.puml
Garma00, line 1: **todo**: generate the png as well inside the `docs/png` folder

apps/pushnotif-func/docs/puml/start-massive-job.puml
```diff
@@ -0,0 +1,169 @@
...
+
+/'
+'  totalTags: Numero totale dei tag. (fisso a livello di function)
+'  groupSize: Numero di tag (batch) per ogni invio singolo. (fisso a livello di function)
+'  executionTime: Delta tra l'invio della prima e dell'ultima notfica in secondi. (ottenuto dal job)
+'
+'  delay: Delta tra l'invio di una batch e la successiva.
+'
+'  delay = executionTime / (totalTags / groupSize)
+' '/
```
Garma00, line 113: **todo**: translate this block if needed

apps/pushnotif-func/docs/puml/start-massive-job.puml
Garma00, line 1: **todo**: generate the png as well inside the `docs/png` folder

apps/pushnotif-func/src/adapters/functions/start-massive-notification-job.ts
```diff
@@ -0,0 +1,60 @@
...
+import {
+  createHttpResponse,
+  parseHttpRequestBody,
+} from "../../domain/http-request";
+import {
+  StartMassiveJobPayloadSchema,
+  massiveJobIDSchema,
+} from "../../domain/massive-jobs";
+
+export const startMassiveNotificationJobHandler =
```
Garma00, line 32: **chore**: always use `get` as prefix here since this is not the handler but a function that returns the handler
```suggestion
export const getStartMassiveNotificationJobHandler =
```

apps/pushnotif-func/src/adapters/functions/start-massive-notification-job.ts
```diff
@@ -0,0 +1,60 @@
...
+  ErrorInternal,
+  ErrorNotFound,
+  ErrorValidation,
+} from "../../domain/error";
+import {
+  createHttpResponse,
+  parseHttpRequestBody,
+} from "../../domain/http-request";
+import {
+  StartMassiveJobPayloadSchema,
```
Garma00: **thought**: I am wondering if it is ok to put the schemas for request validation at domain level.

cc: @giamma1295 @lucamancosu

apps/pushnotif-func/src/adapters/functions/start-massive-notification-job.ts
```diff
@@ -0,0 +1,60 @@
...
+      parsedIdParameter.data,
+      parsedBody.startTimeTimestamp,
+    );
+
+    if (result instanceof ErrorInternal) {
+      return createHttpResponse(500, { error: result.message });
+    }
+    if (result instanceof ErrorNotFound) {
+      return createHttpResponse(404, { error: result.message });
+    }
```
Garma00: **nitpick**: 
```suggestion
    if (result instanceof Error) {
      return createHttpResponse(result.code, { error: result.message });
    }
```

apps/pushnotif-func/src/adapters/storage-queue/check-job-message.ts
```diff
@@ -0,0 +1,31 @@
...
+        {
+          visibilityTimeout: visibilityTimeoutInSeconds,
+        },
+      );
+      return result.messageId;
+    } catch (err) {
+      return new ErrorInternal(
+        "Failed to send message to check job queue",
+        err,
+      );
```
Garma00, line 28: **thought**: I would like to clarify how we can handle those kind of error messages. I think that by simply passing err here we would have `[Object]` as result. Maybe we could use
```ts
JSON.stringify(err)
``` 
but we can never be sure that err is a valid JSON and we could have an unexpected throw, what's your take @giamma1295 @lucamancosu @robbcocco?

apps/pushnotif-func/src/domain/use-cases/create-massive-notification-job.ts
```diff
@@ -3,6 +3,7 @@ import { ulid } from "ulid";
 import { ErrorInternal } from "../error";
 import {
   CreateMassiveJobPayload,
+  MassiveJobResponse,
```
Garma00: **thought**: I am wondering if it is ok to put the schemas for responses at domain level.

apps/pushnotif-func/src/domain/use-cases/start-massive-notification-job.ts
```diff
@@ -0,0 +1,128 @@
...
+
+export class StartMassiveNotificationJobUseCase {
+  private generateAllTags = (n: number): string[] => {
+    const total = 16 ** n;
+    const tags = new Array<string>(total);
+    for (let i = 0; i < total; i++) {
+      tags[i] = i.toString(16).padStart(n, "0");
+    }
+    return tags;
+  };
```
Garma00, line 27: **chore**: add a comment with some context about this method

apps/pushnotif-func/src/domain/use-cases/start-massive-notification-job.ts
```diff
@@ -0,0 +1,128 @@
...
+    if (
+      massiveJob instanceof ErrorInternal ||
+      massiveJob instanceof ErrorNotFound
+    ) {
+      return massiveJob;
+    }
+
+    if (massiveJob.status !== "CREATED") {
+      return new ErrorInternal(
+        `Cannot start massive job with id ${massiveJobId} because it is not in CREATED status`,
```
Garma00: **nitpick**: 
```suggestion
        `Cannot start massive job with id ${massiveJobId} because it is in ${massiveJob.status} status`,
```

apps/pushnotif-func/src/domain/use-cases/start-massive-notification-job.ts
```diff
@@ -0,0 +1,128 @@
...
+    const tags = new Array<string>(total);
+    for (let i = 0; i < total; i++) {
+      tags[i] = i.toString(16).padStart(n, "0");
+    }
+    return tags;
+  };
+
+  constructor(
+    private repository: MassiveJobsRepository,
+    private checkJobMessageQueue: CheckJobMessageQueue,
```
Garma00: **thought**: since we are in the domain here, is it ok to talk about `queue`?

apps/pushnotif-func/src/domain/use-cases/start-massive-notification-job.ts
```diff
@@ -0,0 +1,128 @@
...
+    const updateResult = await this.repository.updateMassiveJob(updatedJob);
+
+    if (
+      updateResult instanceof ErrorInternal ||
+      updateResult instanceof ErrorNotFound
+    ) {
+      return updateResult;
+    }
+
+    const allTags = this.generateAllTags(3); // generates 4096 tags from "000" to "fff"
```
Garma00, line 64: **suggestion**: since those tags are statics we could hardcode them and avoid the computation of generating them.

apps/pushnotif-func/src/domain/use-cases/start-massive-notification-job.ts
```diff
@@ -0,0 +1,128 @@
...
+    const batchSize = 10; // we want to process 10 tags for each batch
+    // we calculate the delay between batches to ensure that all notifications are sent within the expected execution time of the job
+    const delayBetweenBatchesInSeconds =
+      (massiveJob.executionTimeInHours * 3600) / (allTags.length / batchSize);
```
Garma00: **suggestion**: what about adding a numerical example in the comment?

apps/pushnotif-func/src/domain/use-cases/start-massive-notification-job.ts
```diff
@@ -0,0 +1,128 @@
...
+          (1000 * delayBetweenBatchesInSeconds * (index + batchSize)) /
+            batchSize) /
+          1000,
+      );
+
+      const sendNotificationMessage = {
+        jobId: massiveJob.id,
+        scheduledTimestamp,
+        tags,
+      };
```
Garma00, line 88: **suggestion**: this logic is a bit tricky, I would suggest you to put it in a separate method and also add a comment that explain what are we doing here.

apps/pushnotif-func/src/domain/use-cases/start-massive-notification-job.ts
```diff
@@ -0,0 +1,128 @@
...
+        sendNotificationMessages,
+      );
+    } catch (error) {
+      this.telemetryClient.trackException({
+        exception: error instanceof Error ? error : new Error("Unknown error"),
+        properties: {
+          jobId: massiveJob.id,
+        },
+      });
+    }
```
Garma00: **suggestion**: I would prefer to track an event here with a unique name so we can query for it

apps/pushnotif-func/src/domain/massive-jobs.ts
```diff
@@ -67,17 +67,37 @@ export interface MassiveProgressRepository {
...
-  executionTimeInHours: z.number().int().min(2).max(12).default(2),
-  startTimeTimestamp: z
+export const StartMassiveJobPayloadSchema = z.object({
+  startTimeTimestamp: z.coerce
     .number()
     .int()
     .positive()
+    .transform((value) =>
+      value >= 1_000_000_000_000 ? Math.floor(value / 1000) : value,
+    )
```
Garma00: **suggestion**: add a comment that explain this transformation

apps/pushnotif-func/src/domain/send-notification.ts
```diff
@@ -0,0 +1,19 @@
...
+  jobId: z.ulid(),
+  scheduledTimestamp: z.number().int().positive(),
+  tags: z.array(z.string()).default([]),
+});
+
+export type SendNotificationMessage = z.infer<
+  typeof SendNotificationMessageSchema
+>;
+
+export interface SendNotificationMessageQueue {
```
Garma00: **question**: shall this be a repository?

Garma00 (changes requested):

apps/pushnotif-func/docs/puml/create-massive-job.puml
```diff
@@ -0,0 +1,70 @@
...
+activate admin
+activate cj
+
+cj -> cj : Validate request
+
+alt Invalid request
+  cj -[#red]x admin : 400
+end
+
+cj -> cj : Generate jobId
```
lucamancosu: I think JobId is the same Id returned at the end of this flow. Just try to reconcile this nane maybe in the note because is not clear if the returned id is the JobId

apps/pushnotif-func/src/adapters/functions/start-massive-notification-job.ts
giamma1295: @Garma00 @robbcocco I was looking at the current state of the repository and noticed that using ***get*** for these initialization functions results in awkward function names, since it’s often used in HttpHandler names; eg currently we have ***getGet***MassiveNotificationJobHandler. 

What do you think about naming these builder functions to `build<HandlerName>`? 
In the example above, it would become `buildGetMassiveNotificationJobHandler`: simpler and without repetition!

apps/pushnotif-func/src/domain/use-cases/start-massive-notification-job.ts
giamma1295: In addition to what @Garma00 suggested, I’m wondering if using `ErrorInternal` is appropriate in this case; I mean, isn’t it an application error if the request’s job status isn’t “CREATED”? Perhaps an a 4xx code (HTTP code for caller errors) would be more appropriate.

apps/pushnotif-func/src/domain/use-cases/start-massive-notification-job.ts
```diff
@@ -0,0 +1,128 @@
...
+  };
+
+  constructor(
+    private repository: MassiveJobsRepository,
+    private checkJobMessageQueue: CheckJobMessageQueue,
+    private sendNotificationMessageQueueOutput: StorageQueueOutput,
+    private telemetryClient: TelemetryService,
+  ) {}
+
+  async execute(
```
giamma1295: Consider extract some functionalities in private methods, eg `visibilityTimeoutInSeconds` calculation can be done in a dedicated method

apps/pushnotif-func/src/domain/use-cases/start-massive-notification-job.ts
```diff
@@ -0,0 +1,128 @@
+import { InvocationContext, StorageQueueOutput } from "@azure/functions";
```
lucamancosu: This is a strong dependecy to insert inside the domain. I think that a use case must be agnostic of the final implementation and it should only orchestrate domain parts.

My advice is to separate use case and adapter

apps/pushnotif-func/src/domain/use-cases/start-massive-notification-job.ts
lucamancosu: Probably Is better to use domain repos instead

apps/pushnotif-func/src/domain/use-cases/start-massive-notification-job.ts
```diff
@@ -0,0 +1,128 @@
...
+    private sendNotificationMessageQueueOutput: StorageQueueOutput,
+    private telemetryClient: TelemetryService,
+  ) {}
+
+  async execute(
+    context: InvocationContext,
+    massiveJobId: MassiveJobID,
+    startTimeTimestamp: number,
+  ): Promise<ErrorInternal | ErrorNotFound | MassiveJobResponse> {
+    const massiveJob = await this.repository.getMassiveJob(massiveJobId);
```
lucamancosu: If this variable can be an error is better to call it with somenthing like getMassiveJobResponse and call it massive job only if it is not an error

apps/pushnotif-func/src/adapters/functions/start-massive-notification-job.ts
robbcocco: I would avoid this as it would also catch other error types not yet managed. I think something like this would be more explicit:
```
    if (result instanceof ErrorInternal || result instanceof ErrorNotFound) {
      return createHttpResponse(result.code, { error: result.message });
    }
```

apps/pushnotif-func/src/domain/send-notification.ts
robbcocco: Forgot to remove this! Already replaced with an extra output in the endpoint declaration

apps/pushnotif-func/src/adapters/storage-queue/check-job-message.ts
robbcocco: Would something like `err instanceof Error ? JSON.stringify(err) : err` be a good compromise? We'd be sure tu convert errors into strings without the risks of parsing an invalid json altough we'd skip the stringify on non-error objects

apps/pushnotif-func/src/domain/use-cases/start-massive-notification-job.ts
robbcocco: I'd avoid hardcoded values as it could easily introduce errors and reduce scalability if we'd ever need to up to 4 characters. Also as it's run only once per execution, it should not be that heavy on the function.

apps/pushnotif-func/docs/puml/create-massive-job.puml
robbcocco: ab45fbd74542bc3906e4f2b61ce3ffea57261dea

apps/pushnotif-func/docs/puml/start-massive-job.puml
robbcocco: ab45fbd74542bc3906e4f2b61ce3ffea57261dea

apps/pushnotif-func/docs/puml/start-massive-job.puml
robbcocco: ab45fbd74542bc3906e4f2b61ce3ffea57261dea

apps/pushnotif-func/src/adapters/functions/start-massive-notification-job.ts
robbcocco: Went with `make` as the prefix following platform convention in ab45fbd74542bc3906e4f2b61ce3ffea57261dea

apps/pushnotif-func/src/adapters/functions/start-massive-notification-job.ts
robbcocco: ab45fbd74542bc3906e4f2b61ce3ffea57261dea

apps/pushnotif-func/src/domain/use-cases/start-massive-notification-job.ts
robbcocco: ab45fbd74542bc3906e4f2b61ce3ffea57261dea

apps/pushnotif-func/src/domain/use-cases/start-massive-notification-job.ts
robbcocco: ab45fbd74542bc3906e4f2b61ce3ffea57261dea

apps/pushnotif-func/src/domain/use-cases/start-massive-notification-job.ts
robbcocco: ab45fbd74542bc3906e4f2b61ce3ffea57261dea

apps/pushnotif-func/src/domain/massive-jobs.ts
robbcocco: ab45fbd74542bc3906e4f2b61ce3ffea57261dea

apps/pushnotif-func/src/domain/send-notification.ts
robbcocco: Restored in ab45fbd74542bc3906e4f2b61ce3ffea57261dea

apps/pushnotif-func/src/domain/use-cases/start-massive-notification-job.ts
robbcocco: ab45fbd74542bc3906e4f2b61ce3ffea57261dea

apps/pushnotif-func/src/domain/use-cases/start-massive-notification-job.ts
robbcocco: ab45fbd74542bc3906e4f2b61ce3ffea57261dea

Garma00 (approved):
LGTM!

PR #708, opened 2026-04-13
Override attachment_url parameter

Garma00:


Garma00:
Changes already applied locally.

giamma1295 (approved):

PR #712, opened 2026-04-14
Add CancelMassiveJob Function

giamma1295:
Closes IOCOM-3086


apps/pushnotif-func/src/domain/use-cases/cancel-massive-notification-job.ts
```diff
@@ -0,0 +1,112 @@
...
+  constructor(
+    private readonly massiveJobRepository: MassiveJobsRepository,
+    private readonly massiveProgressRepository: MassiveProgressRepository,
+    private readonly pushNotificationRepository: PushNotificationRepository,
+  ) {}
+
+  private async cancelSheduledNotification(
+    progress: MassiveProgress,
+    jobId: MassiveJobID,
+  ): Promise<ErrorInternal | ErrorTooManyRequests | undefined> {
```
Garma00: **suggestion**: return the `notificationID` in the happy path even if we don't use it, I think it is clearer for the happy path.
```suggestion
  ): Promise<ErrorInternal | ErrorTooManyRequests | string> {
```

apps/pushnotif-func/src/domain/use-cases/cancel-massive-notification-job.ts
```diff
@@ -0,0 +1,112 @@
...
+    );
+
+    if (setStatusResult instanceof Error) {
+      return setStatusResult;
+    }
+  }
+
+  private async retrieveJob(
+    jobId: MassiveJobID,
+  ): Promise<ErrorInternal | ErrorNotAccepted | ErrorNotFound | undefined> {
```
Garma00: **issue**: the method is called `retrieveJob` but returns `undefined` in case of success.

**suggestion**: return the retrieved job.

apps/pushnotif-func/src/domain/use-cases/cancel-massive-notification-job.ts
```diff
@@ -0,0 +1,112 @@
...
+  ): Promise<
+    CancelMassiveJobResult | ErrorInternal | ErrorNotAccepted | ErrorNotFound
+  > {
+    const massiveJob = await this.retrieveJob(jobId);
+    if (massiveJob instanceof Error) {
+      return massiveJob;
+    }
+
+    const pendingProgress =
+      await this.massiveProgressRepository.listMassiveJobPendingProgress(jobId);
```
Garma00: **suggestion**: use the id of the retrieved job here and below as well
```suggestion
      await this.massiveProgressRepository.listMassiveJobPendingProgress(massiveJob.id);
```

apps/pushnotif-func/src/domain/massive-jobs.ts
```diff
@@ -9,7 +9,7 @@ export const MassiveProgressStatusEnum = z.enum([
   "PENDING",
   "FAILED",
   "SENT",
-  "CANCELED",
+  "CANCELED", // TODO: this is correct, but NH use the USA spelling "CANCELED", we should check if we want to align with NH or use the correct spelling, this applies to all the codebase
```
Garma00:
```suggestion
  "CANCELED", 
```

Garma00 (changes requested):

apps/pushnotif-func/src/domain/use-cases/cancel-massive-notification-job.ts
giamma1295: Method removed in the last commit!

apps/pushnotif-func/src/domain/use-cases/cancel-massive-notification-job.ts
giamma1295: done in last commit!

apps/pushnotif-func/src/domain/massive-jobs.ts
giamma1295: removed leftover TODO

apps/pushnotif-func/src/domain/use-cases/cancel-massive-notification-job.ts
giamma1295: changed in the last commit, Instead of the notificationID i choosed the status (canceled or failed as "happy path result"  of the cancel operation)

Garma00 (approved):

PR #730, opened 2026-04-21
Pushnotif-func remove azure-storage lib

lucamancosu:
remove azure-storage lib
use MessageContentRepository from io-messages-common-legacy for retrieve the message content
use azure/storage-blob for the blobServices healthcheck

Close IOCOM-2769


Garma00 (approved):
LGTM!

PR #740, opened 2026-04-28
Remove legacy azure-storage lib from services-func

robbcocco:
Closes IOCOM-2767


apps/services-func/src/functions/GetMessage/handler.ts
```diff
@@ -274,7 +274,7 @@ export const GetMessageHandler =
     messageStatusModel: MessageStatusModel,
     notificationModel: NotificationModel,
     notificationStatusModel: NotificationStatusModel,
-    blobService: BlobService,
+    blobService: MessageContentRepository,
```
Garma00:
```suggestion
    messageContentRepository: MessageContentRepository,
```

packages/io-messages-common-legacy/src/domain/message-content.ts
```diff
@@ -1,11 +1,21 @@
 import { MessageContent } from "../types/MessageContent";
 
-export interface MessageContentRepository {
+export interface MessageContentRepository<StoreResponse = unknown> {
```
Garma00: Why are we adding `StoreResponse` as generic here?

apps/services-func/src/functions/GetMessage/handler.ts
```diff
@@ -336,7 +336,18 @@ export const GetMessageHandler =
...
+        TE.tryCatch(
+          () => blobService.getByMessageContentById(retrievedMessage.id),
+          E.toError,
+        ),
+        TE.map((messageContent) => {
+          if (messageContent === null) {
+            return O.none;
+          }
+          return O.some(messageContent);
+        }),
```
Garma00: I prefer this approach cause `O.fromNullable` handles `undefined` as well as `null`.
```suggestion
        TE.map(O.fromNullable),
```

apps/services-func/src/functions/GetMessage/handler.ts
```diff
@@ -496,7 +507,7 @@ export function GetMessage(
   messageStatusModel: MessageStatusModel,
   notificationModel: NotificationModel,
   notificationStatusModel: NotificationStatusModel,
-  blobService: BlobService,
+  blobService: MessageContentRepository,
```
Garma00:
```suggestion
  messageContentRepository: MessageContentRepository,
```

apps/services-func/src/functions/ProcessMessage/handler.ts
```diff
@@ -288,7 +289,7 @@ const createMessageOrThrow = async (
   context: InvocationContext,
   lMessageModel: MessageModel,
   messageStatusUpdater: ReturnType<typeof getMessageStatusUpdater>,
-  lBlobService: BlobService,
+  lBlobService: MessageContentRepository<BlockBlobUploadResponse>,
```
Garma00:
```suggestion
  messageContentRepository: MessageContentRepository<BlockBlobUploadResponse>,
```

apps/services-func/src/functions/ProcessMessage/handler.ts
```diff
@@ -321,13 +322,18 @@ const createMessageOrThrow = async (
...
+        lBlobService.storeMessageContent(newMessageWithoutContent.id, {
+          ...createdMessageEvent.content,
+          payment_data: messagePaymentData,
+        }),
+      E.toError,
+    ),
+    TE.map((messageContent) => {
+      return messageContent;
+    }),
   )();
```
Garma00:
```suggestion
  const errorOrAttachment = await TE.tryCatch(
    () =>
      lBlobService.storeMessageContent(newMessageWithoutContent.id, {
        ...createdMessageEvent.content,
        payment_data: messagePaymentData,
      }),
    E.toError,
  )();

```

apps/services-func/src/functions/ProcessMessage/handler.ts
```diff
@@ -375,7 +381,7 @@ export interface IProcessMessageHandlerInput {
   readonly TTL_FOR_USER_NOT_FOUND: Ttl;
   readonly isOptInEmailEnabled: boolean;
   readonly lActivation: ActivationModel;
-  readonly lBlobService: BlobService;
+  readonly lBlobService: MessageContentRepository<BlockBlobUploadResponse>;
```
Garma00:
```suggestion
  readonly messageCongentRepository: MessageContentRepository<BlockBlobUploadResponse>;
```

apps/services-func/src/utils/with-expanded-input.ts
```diff
@@ -24,28 +24,27 @@ export type DataFetcher<E> = (
...
     blobName: Parameters<typeof getBlobAsObject>[3],
   ): TE.TaskEither<Error, O.Option<A>> =>
-    pipe(
-      TE.tryCatch(
-        () =>
-          getBlobAsObject<A, O, I>(
+    TE.tryCatch(
+      async () =>
+        O.fromNullable(
+          await getBlobAsObject<A, Encoded, I>(
```
Garma00: We should avoid using `await` when dealing with `TaskEither`, I would try with something like this:

```ts
    pipe(
      TE.tryCatch(
        () =>
          getBlobAsObject<A, Encoded, I>(
            type,
            blobService,
            containerName,
            blobName,
            options,
          ),
        E.toError,
      ),
      TE.map(O.fromNullable),
    );

```

packages/io-messages-common-legacy/src/adapters/message-content.ts
```diff
@@ -72,6 +77,28 @@ export class MessageContentBlobAdapter implements MessageContentRepository {
     return response.readableStreamBody;
   }
 
+  async storeMessageContent(
+    messageId: string,
+    content: MessageContent,
+  ): Promise<BlockBlobUploadResponse> {
```
Garma00: The adapter should return a domain type, this way you do not need to add the generic to the entire adapter.

packages/io-messages-common-legacy/src/adapters/message-content.ts
```diff
@@ -72,6 +77,28 @@ export class MessageContentBlobAdapter implements MessageContentRepository {
...
+      .getContainerClient(this.messageContainerName)
+      .getBlockBlobClient(blobName);
+
+    try {
+      return blockBlobClient.upload(
+        JSON.stringify(content),
+        Buffer.byteLength(JSON.stringify(content)),
+      );
+    } catch (e) {
+      if (e instanceof Error) {
```
Garma00: I am not sure but maybe you can use the `isRestError()` function to have the exact error type returned from the storage sdk.

packages/io-messages-common-legacy/src/adapters/utils.ts
```diff
@@ -13,3 +23,89 @@ export const readableStreamToUtf8 = async (
       reject(err);
     });
   });
+
+export const getBlobAsObject = async <A, O, I>(
```
Garma00, line 34: I am not sure but I think we can reuse the `getByMessageContentById` (just noticed the typo) in the message content adapter instead of redefining those functions

Garma00 (changes requested):

apps/services-func/src/functions/CreateMessage/utils.ts
giamma1295, line 1: This whole module can be simplified with something like the following

```ts
import * as E from "fp-ts/lib/Either";
import * as TE from "fp-ts/lib/TaskEither";
import { pipe } from "fp-ts/lib/function";
import { upsertBlobFromObject } from "io-messages-common-legacy/adapters/utils";

export const makeUpsertBlobFromObject =
  (
    blobService: Parameters<typeof upsertBlobFromObject>[0],
    containerName: Parameters<typeof upsertBlobFromObject>[1],
    options: Parameters<typeof upsertBlobFromObject>[4] = {},
  ) =>
  <T>(
    blobName: string,
    content: T,
  ): TE.TaskEither<Error, Awaited<ReturnType<typeof upsertBlobFromObject>>> =>
    pipe(
      TE.tryCatch(
        () =>
          upsertBlobFromObject(
            blobService,
            containerName,
            blobName,
            content,
            options,
          ),
        E.toError,
      ),
    );

```

packages/io-messages-common-legacy/src/adapters/message-content.ts
```diff
@@ -97,4 +102,26 @@ export class MessageContentBlobAdapter implements MessageContentRepository {
...
+        throw new Error(`Cannot store message content as blob: ${e.message}`);
+      }
+      throw new Error(`Cannot store message content as blob: Unknown error`);
+    }
+  }
```
giamma1295, line 119: This Method has 2 problems
1. Double JSON.stringify, so we waste time by doing the same operation twice.
2. Implements somenthing close to `utils.ts upsertBlobFromObject`, so code duplication.

so can be simplified in this way:

```suggestion
  async storeMessageContent(
    messageId: string,
    content: MessageContent,
  ): Promise<BlockBlobUploadResponse> {
    const blobName = `${messageId}.json`;
    try {
      return upsertBlobFromObject(
        this.blobClient,
        this.messageContainerName,
        blobName,
        content,
      );
    } catch (e) {
      if (e instanceof Error) {
        throw new Error(`Cannot store message content as blob: ${e.message}`);
      }
      throw new Error(`Cannot store message content as blob: Unknown error`);
    }
  }
```

apps/services-func/src/main.ts
```diff
@@ -135,13 +136,18 @@ const activationModel = new ActivationModel(
...
 // Blob service for message content storage (reading message body)
-const blobServiceForMessageContent = createBlobService(
-  config.MESSAGE_CONTENT_STORAGE_CONNECTION_STRING,
+const blobServiceClientForMessageContent =
+  BlobServiceClient.fromConnectionString(
+    config.MESSAGE_CONTENT_STORAGE_CONNECTION_STRING,
+  );
+const blobServiceClientForMessageContentAdapter = new MessageContentBlobAdapter(
+  blobServiceClientForMessageContent,
+  "",
```
giamma1295: Empty string instead of container name as constructor argument

giamma1295 (changes requested):

packages/io-messages-common-legacy/src/adapters/message-content.ts
robbcocco: I was discussing this with @lucamancosu earlier, at this moment we do not need any data so we do not have a definition for domain type (hence why i opted for a generic). another alternative could be to return void, since we only check left on the current code, and update the return type if it'll be ever needed

packages/io-messages-common-legacy/src/domain/message-content.ts
robbcocco: See https://github.com/pagopa/io-messages/pull/740#discussion_r3155069438

packages/io-messages-common-legacy/src/adapters/message-content.ts
robbcocco: The [function we're replacing](https://github.com/pagopa/io-functions-commons/blob/85d7f2425bad88f50fe9da6c8a3c5cae47fa4058/src/models/message.ts#L368) did not handle different type of errors, but left this logic to the caller. Do we need to implement such logic?

packages/io-messages-common-legacy/src/adapters/utils.ts
robbcocco: Problem is that getByMessageContentById decodes against MessageContent (hardcoded), while getBlobAsObject uses a generic and is used to decode a CommonMessageData. I think the best thing to do would be to use getBlobAsObject inside getByMessageContentById.

robbcocco: Would also be cleaner and follow the pattern we applied on storeMessageContent, but it's a change that would impact all other projects too.

```
import { MessageContentRepository } from "@/domain/message-content";
import { BlobServiceClient } from "@azure/storage-blob";

import { MessageContent } from "../types/MessageContent";
import { getBlobAsObject, upsertBlobFromObject } from "./utils";

export class BlobStorageErrorException extends Error {
  readonly code: string;

  constructor(code: string, message: string) {
    super(message);
    this.code = code;
    this.name = "BlobStorageErrorException";
  }
}

export class MessageContentBlobAdapter implements MessageContentRepository {
  private readonly blobClient: BlobServiceClient;
  private readonly messageContainerName: string;

  constructor(blobClient: BlobServiceClient, messageContainerName: string) {
    this.blobClient = blobClient;
    this.messageContainerName = messageContainerName;
  }

  async getByMessageContentById(
    messageId: string,
  ): Promise<MessageContent | null> {
    const blobName = `${messageId}.json`;

    try {
      return getBlobAsObject(
        MessageContent,
        this.blobClient,
        this.messageContainerName,
        blobName,
      );
    } catch (e) {
      throw new Error(
        `Cannot parse content text into object: ${e instanceof Error ? e.message : "Unknown error"}`,
      );
    }
  }

  async storeMessageContent(
    messageId: string,
    content: MessageContent,
  ): Promise<void> {
    const blobName = `${messageId}.json`;

    try {
      await upsertBlobFromObject(
        this.blobClient,
        this.messageContainerName,
        blobName,
        content,
      );
    } catch (e) {
      throw new Error(
        `Cannot parse content text into object: ${e instanceof Error ? e.message : "Unknown error"}`,
      );
    }
  }
}

```

apps/services-func/src/functions/GetMessage/handler.ts
robbcocco: Resolved in 211de0895ff7bb39c81f30cd222f360895012887

apps/services-func/src/functions/GetMessage/handler.ts
robbcocco: Resolved in 211de0895ff7bb39c81f30cd222f360895012887

apps/services-func/src/functions/GetMessage/handler.ts
robbcocco: Resolved in 211de0895ff7bb39c81f30cd222f360895012887

apps/services-func/src/functions/ProcessMessage/handler.ts
robbcocco: Resolved in 211de0895ff7bb39c81f30cd222f360895012887

apps/services-func/src/functions/ProcessMessage/handler.ts
robbcocco: Resolved in 211de0895ff7bb39c81f30cd222f360895012887

apps/services-func/src/functions/CreateMessage/utils.ts
robbcocco: Resolved in 211de0895ff7bb39c81f30cd222f360895012887

packages/io-messages-common-legacy/src/adapters/message-content.ts
robbcocco: Resolved in 211de0895ff7bb39c81f30cd222f360895012887

apps/services-func/src/main.ts
robbcocco: Resolved in 211de0895ff7bb39c81f30cd222f360895012887

packages/io-messages-common-legacy/src/adapters/message-content.ts
```diff
@@ -97,4 +97,24 @@ export class MessageContentBlobAdapter implements MessageContentRepository {
...
+    try {
+      await upsertBlobFromObject(
+        this.blobClient,
+        this.messageContainerName,
+        blobName,
+        content,
+      );
+    } catch (e) {
+      throw new Error(
+        `Cannot parse content text into object: ${e instanceof Error ? e.message : "Unknown error"}`,
```
Garma00: What about adding the blob name in the error message?

apps/services-func/package.json
giamma1295: Removal of this dependency should be the task goal

giamma1295 (changes requested):

Garma00 (approved):

giamma1295 (approved):

packages/io-messages-common-legacy/src/adapters/message-content.ts
robbcocco: Resolved in ec861fb

apps/services-func/package.json
robbcocco: Resolved in ec861fb

PR #748, opened 2026-05-07
Handle race condition in start massive jobs

robbcocco:


apps/pushnotif-func/docs/puml/start-massive-job.puml
```diff
@@ -115,7 +115,10 @@ activate mnj
 '  delay = executionTime / (totalTags / groupSize)
 ' '/
 
-alt Error from comos
+alt Concurrent modification detected
+  mnj -[#red]> httpfunc : Update rejected due to ETag mismatch
+  httpfunc -[#red]x admin : 412
+else Error from cosmos
```
Garma00, line 121: In the sequence the function returns `412` but actually returns `500`

Garma00 (changes requested):

apps/pushnotif-func/docs/puml/start-massive-job.puml
robbcocco: Resolved in b1fb3a3

Garma00 (approved):

PR #763, opened 2026-05-22
Add minimum release age in package.json

Garma00:


package.json
```diff
@@ -26,5 +26,6 @@
     "typescript": "catalog:",
     "vitest": "catalog:"
   },
+  "minimumReleaseAge": 1440,
```
giamma1295, line 29: We may want a whitelist for internal libs? if so we can configure the MinimumReleaseAge on `pnpm-workspace.yaml` which support whitelists!
https://pnpm.io/it/settings#minimumreleaseage

```yaml
minimumReleaseAge: 1440
minimumReleaseAgeExclude:
  - '@pagopa/*'
 ```

Garma00: I would prefer to avoid this exclude until it's needed simply because a dependency under `@pagopa/` could be "infected" cause of a supply chain attack, what's your take?

giamma1295: Yes, it is possible that the internal library could be compromised, although we usually release new versions of internal libraries as needed by the projects that use them!
If we want to be 100% safe while remaining flexible, we could update the `package.json` (or `pnpm-workspace.yaml`) files for the internal libraries as well!
Anyway we can go with this atm!

giamma1295 (approved):

PR #767, opened 2026-05-26
Feature Flag to gate massive job functions mounting

daniacca:
This PR introduces an environment flag in the `pushnotif` service to manually enable or disable the mounting of massive job functions. These functions cannot currently be used due to external blockers. When disabled, the new flag prevents those functions from being loaded by the service, reducing resource consumption.

Closes IOCOM-3150


apps/pushnotif-func/src/adapters/config.ts
```diff
@@ -153,6 +162,7 @@ const mapEnvironmentVariablesToConfig = (env: Env): Config => ({
   },
   comStorageConnectionString: env.NOTIFICATIONS_STORAGE_CONNECTION_STRING,
   databaseName: env.PUSH_DATABASE_NAME,
+  enableMassiveNotificationJobs: env.ENABLE_MASSIVE_NOTIFICATION_JOBS ?? false,
```
Garma00: **Suggestion**: Directly use the aod `default` inside the type definition
```suggestion
  enableMassiveNotificationJobs: env.ENABLE_MASSIVE_NOTIFICATION_JOBS,
```

apps/pushnotif-func/src/adapters/config.ts
```diff
@@ -67,9 +67,17 @@ const envSchema = z.object({
...
 
   COSMOSDB_URI: z.url(),
+  ENABLE_MASSIVE_NOTIFICATION_JOBS: z
+    .string()
+    .transform((val) => {
+      const sanitizedVal = val.trim().toLowerCase();
+      return sanitizedVal === "true" || sanitizedVal === "1";
+    })
+    .pipe(z.boolean())
+    .optional(),
```
Garma00: See comment below
```suggestion
    .default("false"),
```

Garma00 (approved):
Well done! I just left a small improvement in he comments below.

apps/pushnotif-func/src/adapters/config.ts
daniacca: I was torn between leaving it optional or with a default false, but maybe you're right: with the default value we still have safety if the variable is not provided but with better readability!

PR #777, opened 2026-05-29
Add create notification sequence diagram

lucamancosu:
Closes IOCOM-3176


apps/services-func/docs/puml/create-notification.puml
```diff
@@ -0,0 +1,189 @@
...
+
+deactivate procblob
+
+== Email channel decision ==
+
+note right of func
+  **isEmailBlockedForService**
+  Only applies when servicePreferencesSettings.mode
+  is LEGACY and EMAIL is in blockedInboxOrChannels.
+end note
```
Garma00, line 88: It is not clear what the `blockedInboxOrChannels` represents. I would prefer to add some context in this note instead of using the variable's names that are in the code.

I would try to answer those 2 questions here:

- What does it means that `ervicePreferencesSettings.mode` is `LEGACY`?
-  What does it means that `blockedInboxOrChannels` is `true`?

apps/services-func/docs/puml/create-notification.puml
```diff
@@ -0,0 +1,189 @@
...
+
+func -> func : Compute isWebhookEnabledInProfile\n(profile.isWebhookEnabled === true)
+
+func -> func : Compute isWebhookBlockedForService\n(WEBHOOK in blockedInboxOrChannels)
+
+note right of func
+  Webhook notification is sent only when:
+  - isWebhookEnabledInProfile
+  - !isWebhookBlockedForService
+end note
```
Garma00: Ditto. Take the following refactor as an example:

```suggestion
note right of func
  Webhook notification is sent only when:
  - Recipient has "receive webhook notifications" enabled.
  - Recipient has "receive webhook notifications" enabled for the sender service.
end note
```

I would avoid to add lots of variables name.

apps/services-func/docs/puml/create-notification.puml
```diff
@@ -0,0 +1,189 @@
...
+func -> func : Resolve maybeWebhookNotificationUrl\n(O.none if any condition fails)
```
Garma00: Ditto. `O.none` is a concept only a few developers understand.

Garma00 (changes requested):

Garma00 (approved):
LGTM!

PR #778, opened 2026-06-03
Webhook notification sequence diagram

daniacca:
Close IOCOM-3178


apps/services-func/docs/puml/webhook-notification.puml
```diff
@@ -0,0 +1,153 @@
...
+
+func -> notifdb : Find notification by\n(id=notificationId, pk=messageId)
+activate notifdb
+
+alt Error from Cosmos
+  notifdb -[#red]> func
+  func -[#red]x inq : throw Error (retry)
+else Notification not found
+  notifdb -[#red]> func : Empty result
+  func -[#red]x inq : throw Error (retry)
```
Garma00: **Suggestion**: You can unify those 2 cases since the behaviour is the same, something like [this](https://github.com/pagopa/io-messages/blob/main/apps/citizen-func/docs/puml/upsert-message-status.puml#L94).

apps/services-func/docs/puml/webhook-notification.puml
```diff
@@ -0,0 +1,153 @@
...
+else Notification found
+  notifdb -[#green]> func : Notification document
```
Garma00, line 93: **Suggestion**: You can add a note with an example response from the database, this can help the reader to better understand what informations are retrieved and the shape that they have. See [this](https://github.com/pagopa/io-messages/blob/main/apps/citizen-func/docs/puml/upsert-message-status.puml#L103) for example.

apps/services-func/docs/puml/webhook-notification.puml
```diff
@@ -0,0 +1,153 @@
...
+deactivate procblob
+
+func -> func : Check message validity\n(ActiveMessage.decode)
+
+alt Message expired (TTL)
+  func -> func : Log RESULT=TTL_EXPIRED
+  func -[#green]> inq : Return SUCCESS / EXPIRED
+end
+
+func -> notifdb : Find notification by\n(id=notificationId, pk=messageId)
```
Garma00, line 68: It seems like we ask for the notification to the database but then we never use it in this function, interesting. I think we can remove this part in the future (from the diagram and the code as well).

apps/services-func/docs/puml/webhook-notification.puml
```diff
@@ -0,0 +1,153 @@
...
+  is NOT used as the HTTP target.
+  `webhookEndpoint` is sent in the body but is not part of
+  pushnotif-func's NotificationInfo schema (effectively ignored).
+end note
+
+alt Timeout (AbortError)
+  notify -[#red]> func : AbortError
+  func -> func : Map to TransientError
+  func -[#red]x inq : throw Error (retry)
+else Other fetch exception (network/TLS/DNS)
```
Garma00, line 122: Interesting that we don't retry the operation in this case, I think we can discuss this in future!

apps/services-func/docs/puml/webhook-notification.puml
```diff
@@ -0,0 +1,153 @@
...
+  func -> func : Map to PermanentError
+  func -> func : Log ERROR=<message>
+  func -[#orange]> inq : Return FAILURE / SEND_TO_WEBHOOK_FAILED
+else HTTP 204
+  notify -[#green]> func : 204 No Content
+  ' Out-of-scope for services-func; shown for context only.
+  notify --> nh : (assumed) enqueue NotifyMessage -> NH dispatch
+  nh --> webhook : (assumed) APNs / FCM push
+  webhook --> nh : (assumed) delivery ack
+  func -[#green]> inq : Return SUCCESS / OK
```
Garma00: **Suggestion**: Since this is out of scope (as you say in the comment) I would remove this block. We will have a separate and dedicated sequence diagram that will describe this flow.

Garma00 (approved):

PR #789, opened 2026-06-09
Remove connection string from citizen-func

daniacca:
Partial PR for IOCOM-3194, only for `citizen-func`

### DIFF
--- a/apps/citizen-func/src/main.ts
+++ b/apps/citizen-func/src/main.ts
@@ -1,4 +1,5 @@
 import { app } from "@azure/functions";
+import { DefaultAzureCredential } from "@azure/identity";
 import { BlobServiceClient } from "@azure/storage-blob";
 import {
   MESSAGE_COLLECTION_NAME,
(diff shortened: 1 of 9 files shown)


apps/citizen-func/src/main.ts
```diff
@@ -65,10 +66,17 @@ const messageViewModel = new MessageViewExtendedQueryModel(
   cosmosdbInstance.container(MESSAGE_VIEW_COLLECTION_NAME),
 );
 
+const aadCredentials = new DefaultAzureCredential();
+
 const messageContentRepository = new MessageContentBlobAdapter(
-  BlobServiceClient.fromConnectionString(
-    config.MESSAGE_CONTENT_STORAGE_CONNECTION_STRING,
-  ),
+  config.isProduction
```
lucamancosu: This looks similar to the code inside the healthcheck. Can we pass pass the same messageContentRepository to the info handler ? (or create and pass a factory if the repository is not needed on the healthcheck)

apps/citizen-func/src/utils/config.ts
```diff
@@ -65,37 +67,48 @@ export const UlidMapFromString = new t.Type<ReadonlyMap<string, Ulid>, string>(
...
-    isProduction: t.boolean,
-    /* eslint-enable sort-keys */
-  }),
+const IStorageAccountConnectionString = t.type({
+  MESSAGE_CONTENT_STORAGE_CONNECTION_STRING: NonEmptyString,
+  isProduction: t.literal(false),
+});
+
+const IStorageAccountEndpoint = t.type({
+  MESSAGE_CONTENT_STORAGE_ENDPOINT: NonEmptyString,
```
lucamancosu, line 103: Please add it to the local.settings.example

daniacca: We choose not to add this in the local.settings since for local development the key `MESSAGE_CONTENT_STORAGE_ENDPOINT` should not be used (it is also not necessary for the configuration)

apps/citizen-func/src/main.ts
daniacca: 61cc7597df042bac51af7542dadc8ff18817e781

lucamancosu (approved):
LGTM. Gj

Garma00 (approved):

PR #798, opened 2026-06-11
Removed Api Storage Account Connection string from pushnotif-func main

daniacca:
Removed connection string connection on Azure Blob Storage for message content on the `pushnotif-func`.

work in progress for IOCOM-3194

### DIFF
--- a/apps/pushnotif-func/src/adapters/config.ts
+++ b/apps/pushnotif-func/src/adapters/config.ts
@@ -93,8 +93,6 @@ const basicSchema = z.object({
   NH4_NAME: z.string().min(1),
   NH4_PARTITION_REGEX: z.string().min(1),
 
-  NOTIFICATIONS_STORAGE_CONNECTION_STRING: z.string().min(1),
-
   PUSH_DATABASE_NAME: z.string().min(1),
 
   SESSION_MANAGER_API_KEY: z.string().min(1),
@@ -109,11 +107,13 @@ const basicSchema = z.object({
 const contentStorageConnectionStringSchema = z.object({
   MESSAGE_CONTENT_STORAGE_CONNECTION_STRING: z.string().min(1),
   NODE_ENV: z.literal("development"),
+  NOTIFICATIONS_STORAGE_CONNECTION_STRING: z.string().min(1),
 });
 
 const contentStorageEndpointSchema = z.object({
   MESSAGE_CONTENT_STORAGE_ENDPOINT: z.url(),
   NODE_ENV: z.literal("production"),
+  NOTIFICATIONS_STORAGE_QUEUE_ENDPOINT: z.url(),
 });
 
 const envSchema = z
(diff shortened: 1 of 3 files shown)


apps/pushnotif-func/src/main.ts
giamma1295: Async trigger needs to be updated as well in order to auth using ManagedIdentities!

giamma1295 (changes requested):

apps/pushnotif-func/src/main.ts
daniacca: I've leaved this since in `development` we still using the connection string.
When we remove the string from `production` env, I've defined the `NOTIFICATIONS_STORAGE_CONNECTION_STRING__queueServiceUri` on Terraform which should be used by default from Az for identity-based storage connections.

The same things is already done on `ops-func`, you can find the `STORAGE_ACCOUNT__queueServiceUri` under config.ts (`ops-func/src/adapter/config`) and used on line 61 on `ops-func/src/func`

giamma1295: yup it works!, maybe we can change the name to a more generic one (adding some comment explaining how to configure), this is too specific for connectionString.
What about something like `NOTIFICATIONS_STORAGE_CONNECTION` ?

daniacca: good point, I'm on it :)

apps/pushnotif-func/src/adapters/config.ts
```diff
@@ -113,28 +104,44 @@ const envSchema = z.object({
...
 
+const contentStorageConnectionStringSchema = z.object({
+  MESSAGE_CONTENT_STORAGE_CONNECTION_STRING: z.string().min(1),
+  NODE_ENV: z.literal("development"),
+  NOTIFICATIONS_STORAGE_CONNECTION_STRING: z.string().min(1),
+});
+
+const contentStorageEndpointSchema = z.object({
+  MESSAGE_CONTENT_STORAGE_ENDPOINT: z.url(),
+  NODE_ENV: z.literal("production"),
```
Garma00, line 115: I think it is better to leave `NODE_ENV` in the basic config, it is not a variable related to the storage.

Garma00 (approved):

apps/pushnotif-func/src/adapters/config.ts
daniacca: I initially leaved the `NODE_ENV` in the basic config, but in the end moved there because it's the discriminant literal over the discriminated union. I choose to not introduce a new variable in the config for this discrimination and use this directly!

apps/pushnotif-func/src/main.ts
daniacca: f3a0d3e2efa9e84d96cd252f8f2c237c7c87d185

giamma1295 (approved):
LGTM

PR #805, opened 2026-06-15
Update cqrs func for removing blob connection string

daniacca:
Removing all the connection string for blob storage for `cqrs-func`
WPI for IOCOM-3194

### DIFF
--- a/apps/cqrs-func/src/main.ts
+++ b/apps/cqrs-func/src/main.ts
@@ -23,9 +23,7 @@ const config = getConfigOrThrow();
 
 const telemetryClient = initTelemetryClient(config);
 
-const aadCredentials = config.isProduction
-  ? new DefaultAzureCredential()
-  : undefined;
+const aadCredentials = new DefaultAzureCredential()
 
 const kafkaMessagesClient = fromSas(
   config.MESSAGES_TOPIC_CONNECTION_STRING,
--- a/apps/cqrs-func/src/main.ts
+++ b/apps/cqrs-func/src/main.ts
@@ -23,7 +23,7 @@ const config = getConfigOrThrow();
 
 const telemetryClient = initTelemetryClient(config);
 
-const aadCredentials = new DefaultAzureCredential()
+const aadCredentials = new DefaultAzureCredential();
 
 const kafkaMessagesClient = fromSas(
   config.MESSAGES_TOPIC_CONNECTION_STRING,
--- a/apps/cqrs-func/src/main.ts
+++ b/apps/cqrs-func/src/main.ts
@@ -49,15 +49,17 @@ const errorStorage = config.isProduction
       config.MESSAGE_PAYMENT_UPDATER_FAILURE_QUEUE_NAME,
     );
 
+const messageContentBlobClient = config.isProduction
+  ? new BlobServiceClient(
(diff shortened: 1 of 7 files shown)


infra/resources/_modules/web_apps/cqrs-func.tf
```diff
@@ -44,6 +44,10 @@ locals {
       FETCH_KEEPALIVE_TIMEOUT             = "60000"
 
       COM_STORAGE_CONNECTION_STRING = var.com_st_connectiostring
+
+      // Used by storage queue connection, the queueServiceUri is used in production environment
+      COM_STORAGE_CONNECTION_STRING__queueServiceUri = var.com_st_queue_uri
```
giamma1295: i suggest to call this in a more generic way, this is too specific for connection string!

daniacca: ops,we already did the same on citizen, missed this one!

apps/cqrs-func/src/utils/healthcheck.ts
```diff
@@ -85,12 +84,12 @@ export const checkAzureCosmosDbHealth = (
 /**
  * Check the application can connect to an Azure Storage
  *
- * @param connStr connection string for the storage
+ * @param blobClient BlobServiceClient instance
  *
  * @returns either true or an array of error messages
  */
 export const checkAzureStorageHealth = (
```
giamma1295, line 90: You can simplify this function in this way(with this you can remove the `fp-ts/lib/Array` import as well)

```typescript
export const checkAzureStorageHealth = (
  blobClient: BlobServiceClient,
): HealthCheck<"AzureStorage", true> =>
  pipe(
    TE.tryCatch(
      () => blobClient.getProperties(),
      toHealthProblems("AzureStorage"),
    ),
    TE.map((_) => true),
  );
```

infra/resources/_modules/web_apps/cqrs-func.tf
daniacca: 59fd90aa64616c49a3a7be7e2ec81c73e30678c7

apps/cqrs-func/src/utils/healthcheck.ts
daniacca: f313cd354f36a08f9c787c05b15184ed2f652a17

daniacca: You're right, originally the healthcheck was checking on 2 condition simultaneusly, so the use of the array, but as we already did on citizen, we simplified the healthcheck to only one condition and there is need for the array check anymore!

apps/cqrs-func/src/main.ts
```diff
@@ -22,6 +23,10 @@ const config = getConfigOrThrow();
 
 const telemetryClient = initTelemetryClient(config);
 
+const aadCredentials = config.isProduction
+  ? new DefaultAzureCredential()
+  : undefined;
```
Garma00: I would call `new DefaultAzureCredential()` anyway as we did for the `citizen-func`
```suggestion
const aadCredentials = new DefaultAzureCredential()
```

Garma00 (changes requested):

Garma00 (approved):

PR #820, opened 2026-06-23
Upgrade terraform to v1.15.6

Garma00:


giamma1295 (approved):
LGTM,
However, we should update the Terraform version in the devcontainer (which was older than the version in the target repository; it was using 1.10.1).

PR #827, opened 2026-06-30
Add messages app

Garma00:
Add a new fastify app that will handle all the logic about IO Messages.

Resolves IOCOM-3222


apps/messages-app/src/adapters/outbound/package-json/package-json-app-info-reader.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    // Once we ensured that the JSON is valid, we parse it into `AppInfo`.
+    const appInfoParseResult = safeParseJSONToAppInfo(parseResult.value);
+    if (appInfoParseResult.isErr()) {
+      return err(appInfoParseResult.error);
+    }
+
+    const appInfo = appInfoParseResult.value;
+
+    return ok(appInfo);
```
daniacca: this is working, but we are accessing the file system (slow operation) and we are using and hardcoded relative path to retrieve the file location.

Perhaps, we can simplify this adapter, taking advantages of [npm package json vars](https://docs.npmjs.com/cli/v6/using-npm/scripts#packagejson-vars) and access the version information directly by reading the process env var, like: `process.env.npm_package_version`.

This will work only if the application is launched with `npm` or similar command is used to launch the application, but then we will have all the relevant information already available in memory and there is no need to open and read the file. What do you think?

Garma00: @daniacca I understand your point about the slow operation, but I'd prefer to avoid adding a new dependency just for this.

I was thinking we could cache the information using a singleton approach: we read the `package.json` once and store the data in memory (perhaps as a static class property). What do you think?

daniacca: Actually, for accessing the env var via `process` I don't think we need any dependency

lucamancosu (approved):
LGTM. Gj!

apps/messages-app/src/adapters/outbound/package-json/package-json-app-info-reader.ts
Garma00: Sorry I misunderstood, your approach is quite good.

[9de55e9](https://github.com/pagopa/io-messages/pull/827/commits/9de55e9a8b7d6ad85a3095cab2601e16591f5e18)

PR #835, opened 2026-07-07
Add healthcheck endpoint to messages-app

robbcocco:
Closes IOCOM-3235

### DIFF
--- a/apps/messages-app/src/adapters/outbound/healthcheckers/cosmos.adapter.ts
+++ b/apps/messages-app/src/adapters/outbound/healthcheckers/cosmos.adapter.ts
@@ -1,0 +1,28 @@
+import { AppHealthchecker } from "@/application/ports/app-healthcheck.js";
+import { CosmosClient, RestError } from "@azure/cosmos";
+import { GenericError } from "@pagopa/hexagonal-core";
+import { Result, err, ok } from "neverthrow";
+
+export class CosmosClientHealthcheckAdapter implements AppHealthchecker {
+  constructor(private cosmosClient: CosmosClient) {}
+
+  async health(): Promise<Result<void, GenericError>> {
+    try {
+      await this.cosmosClient.databases.readAll().fetchAll();
+      return ok(undefined);
+    } catch (e) {
+      if (e instanceof RestError) {
+        return err(
+          new GenericError(
+            `cosmos db unavailable: ${e.statusCode} ${e.message}`,
+          ),
+        );
+      }
+      return err(
+        new GenericError(
+          `unexpected error: ${e instanceof Error ? e.message : String(e)}`,
+        ),
+      );
+    }
+  }
+}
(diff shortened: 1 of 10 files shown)


lucamancosu (approved):
LGTM

apps/messages-app/src/adapters/outbound/healthcheckers/cosmos.adapter.ts
```diff
@@ -0,0 +1,28 @@
...
+import { CosmosClient, RestError } from "@azure/cosmos";
+import { GenericError } from "@pagopa/hexagonal-core";
+import { Result, err, ok } from "neverthrow";
+
+export class CosmosClientHealthcheckAdapter implements AppHealthchecker {
+  constructor(private cosmosClient: CosmosClient) {}
+
+  async health(): Promise<Result<void, GenericError>> {
+    try {
+      await this.cosmosClient.databases.readAll().fetchAll();
```
Garma00: This operation requires permissions at the Cosmos DB account scope. 

Currently, the application adheres to the principle of least privilege: it is granted access exclusively to the 2 specific containers it needs (see [here](https://github.com/pagopa/io-messages/blob/main/infra/resources/_modules/container_apps/messages.tf#L73-L73)).

I suggest using an alternative health check method. For example, we can perform a point read on a purposefully non-existent item within one of our accessible containers. A `404 Not Found` response will safely confirm that the database is up, reachable, and correctly authorizing our identity.

Something like this:

```ts
const { statusCode } = await this.cosmosClient
      .database("foo")
      .container("bar")
      .item("healthcheck-test", "healthcheck-test")
      .read();

if (statusCode === 200 || statusCode === 404) {
      return { status: 'up', message: 'Cosmos DB is reachable' };
}
```

What's your take? cc: @giamma1295 @lucamancosu @daniacca

apps/messages-app/src/adapters/outbound/healthcheckers/storage-blob.adapter.ts
```diff
@@ -0,0 +1,24 @@
...
+import { BlobServiceClient, RestError } from "@azure/storage-blob";
+import { GenericError } from "@pagopa/hexagonal-core";
+import { Result, err, ok } from "neverthrow";
+
+export class StorageBlobHealthcheckAdapter implements AppHealthchecker {
+  constructor(private blobServiceClient: BlobServiceClient) {}
+
+  async health(): Promise<Result<void, GenericError>> {
+    try {
+      await this.blobServiceClient.getProperties();
```
Garma00: Ditto

Garma00 (changes requested):

apps/messages-app/src/adapters/outbound/healthcheckers/cosmos.adapter.ts
robbcocco: Resolved in cfd1f11

apps/messages-app/src/adapters/outbound/healthcheckers/storage-blob.adapter.ts
robbcocco: Resolved in cfd1f11

lucamancosu (approved):

Garma00 (approved):

PR #839, opened 2026-07-10
Add get messages endpoint to messages-app

Garma00:
- Add the get messages endpoint to messages app
- Add cosmos local emulator for testing it
- Add app settings

Closes IOCOM-3236


infra/resources/_modules/container_apps/messages.tf
```diff
@@ -21,9 +21,16 @@ module "messages_ca" {
...
-        PORT     = 3000
+        HOST                            = "0.0.0.0"
+        NODE_ENV                        = "production"
+        PORT                            = 3000
+        COMMON_COSMOS_DATABASE_NAME     = "db"
+        MESSAGE_METADATA_CONTAINER_NAME = "messages"
+        MESSAGE_STATUS_CONTAINER_NAME   = "message-status"
+        MESSAGE_CONTENT_CONTAINER_NAME  = "message-content"
+        COMMON_COSMOS_URI               = var.common_cosmos_account.endpoint
+        COMMON_STORAGE_ACCOUNT_URI      = var.common_storage_account.endpoint
```
giamma1295, line 32: Just to check, are role to the new Messages Container App, for both cosmos and storage account, assigned?

Garma00: Nice catch, I think we need to add them, I'll check it today.

apps/messages-app/src/application/use-cases/get-user-messages.use-case.ts
```diff
@@ -0,0 +1,273 @@
...
+      }
+
+      const batch = metadata.value;
+
+      // If we received fewer metadatas than the requested buffer, there are no
+      // more messages to retrieve: we reached the end of the page. This also
+      // avoids an extra empty round-trip to Cosmos just to discover the end.
+      if (batch.length < pageSize * 2) {
+        endOfPageReached = true;
+      }
```
giamma1295: `message-metadata-repository` filter's out eventually malformed items:

```typescript
const decodedMetadatas: MessageMetadata[] = [];
    for (const resource of resources.resources) {
      const parsed = cosmosMessageMetadataSchema.safeParse(resource);
      if (parsed.success) {
        decodedMetadatas.push(toMessageMetadata(parsed.data));
      }
    }
```
in this way ewe don't risk skipping pages? or am i missing something?

apps/messages-app/src/application/use-cases/get-user-messages.use-case.ts
```diff
@@ -0,0 +1,273 @@
...
+          // every iteration. Unlike the `cursor` (upper bound) it does not
+          // advance, it only constrains the result to messages newer than it.
+          minimumID,
+        );
+
+      if (metadata.isErr()) {
+        return err(metadata.error);
+      }
+
+      const batch = metadata.value;
```
giamma1295: nit-pick: To make the code more readable, consider using a more descriptive variable name.
For example, 
batch -> messageMetadataPage(Batch)
metadata -> messageMetadataResponse

apps/messages-app/src/application/use-cases/get-user-messages.use-case.ts
```diff
@@ -0,0 +1,273 @@
...
+
+      const metadataById = new Map(batch.map((m) => [m.id, m]));
+
+      for (const status of statuses.value) {
+        // We only keep the messages matching the requested archived flag.
+        if (status.isArchived !== archived) {
+          continue;
+        }
+
+        const metadata = metadataById.get(status.messageId);
```
giamma1295: metadata const name was previously used into the parent scope(row 146)

apps/messages-app/src/application/use-cases/get-user-messages.use-case.ts
Garma00: Yes you are right. 
On the other hand the current implementation returns `500` if there is a malformed item so I think this is not a regretion but an improvement. What's your take? @giamma1295

apps/messages-app/src/application/use-cases/get-user-messages.use-case.ts
```diff
@@ -0,0 +1,273 @@
...
+      await messageContentRepository.getMessagesContentByIds(
+        selectedMessages.map((s) => s.metadata.id),
+      );
+
+    if (messageContents.isErr()) {
+      return err(messageContents.error);
+    }
+
+    const contentById = messageContents.value;
+    const items: PublicMessage[] = [];
```
giamma1295, line 225: Consider using a more descriptive const name instead of generic items to improve code readability, something like `enrichedMessages` should do the trick!

apps/messages-app/src/application/use-cases/get-user-messages.use-case.ts
```diff
@@ -0,0 +1,273 @@
...
+    //
+    // - `prev` is the id of the first (newest) message of the page and it is
+    //   used to navigate to the previous (more recent) page.
+    // - `next` is the id of the last (oldest) message of the page and it is set
+    //   only when there are more messages to retrieve.
+    return ok({
+      items,
+      next: hasMoreResults ? items.at(-1)?.id : undefined,
+      prev: items.at(0)?.id,
+    });
```
giamma1295, line 272: next should contains always the oldest message id on the current message page, actually this is not always true, cause if can be skipped on the above messages enrichment for loop, better calculate next using the not enriched message page.

Edge-Case: If no messages are enriched, `next` will be `undefined` and `items` will be an empty list; this will signal to the client that no messages were found for the user and that it should not attempt to retrieve any more messages—which is not actually true, but is merely due to a problem with the enrichments

giamma1295 (commented):
overall seems ok, only few comments!

apps/messages-app/src/application/use-cases/get-user-messages.use-case.ts
Garma00: [cac81dd](https://github.com/pagopa/io-messages/pull/839/commits/cac81ddfca133488950074fc4c1251f6bad3bea1)

apps/messages-app/src/application/use-cases/get-user-messages.use-case.ts
Garma00: [30b0f9a](https://github.com/pagopa/io-messages/pull/839/commits/30b0f9adc8ae2ccd5c1da63a0eb71fafd258046d)

apps/messages-app/src/application/use-cases/get-user-messages.use-case.ts
Garma00: [30b0f9a](https://github.com/pagopa/io-messages/pull/839/commits/30b0f9adc8ae2ccd5c1da63a0eb71fafd258046d)

lucamancosu (approved):

PR #850, opened 2026-07-15
Add MalformedEntity error

Garma00:
- Add MalformedEntityError
- Replace ValidationError with MalformedEntityError
- Add `.prettierignore`

Closes IOCOM-3266


giamma1295 (approved):
LGTM, in future PR we should add some test cause the codebase on this new function is growing really fast!

PR #854, opened 2026-07-15
Add logs in messages app

robbcocco:
- Update `@pagopa/hexagonal-core` to 0.4.0
- Update `@pagopa/hexagonal-fastify` to 0.2.0
- Add `@opentelemetry/api-logs` and `@pagopa/azure-tracing`

Closes IOCOM-3265


pnpm-workspace.yaml
```diff
@@ -17,10 +17,13 @@ catalog:
...
   "@azure/storage-blob": ^12.29.1
   "@azure/storage-queue": ^12.28.1
+  "@opentelemetry/api-logs": ^0.218.0
+  "@pagopa/azure-tracing": ^0.5.5
   "@pagopa/eslint-config": ^4.0.6
   "@pagopa/fp-ts-kafkajs": ^0.3.0
-  "@pagopa/hexagonal-core": ^0.1.1
+  "@pagopa/hexagonal-core": ^0.4.0
   "@pagopa/hexagonal-fastify": ^0.1.0
+  "@pagopa/hexagonal-openapi": ^0.1.1
```
giamma1295: Actually there is no need to import the `@pagopa/hexagonal-openapi` lib, to solve issue with the RouteContract after updating the `@pagopa/hexagonal-core` to v0.4.0:
- remove "operationId" on all defineRoute
- update `@pagopa/hexagonal-fastify` to v0.2.0

giamma1295 (changes requested):

apps/messages-app/src/adapters/outbound/message/message-metadata.adapter.ts
```diff
@@ -119,13 +122,22 @@ export class MessageMetadataCosmosAdapter implements MessageMetadataRepository {
...
         decodedMetadatas.push(toMessageMetadata(parsed.data));
+      } else {
+        this.logger.trackEvent({
+          name: "MessageMetadataCosmosAdapter.getMessagesMetadataByUser.failed.parse",
+          properties: {
+            fiscalCode,
+            maximumID: maximumID ? maximumID : "undefined",
+            minimumID: minimumID ? minimumID : "undefined",
+            pageSize,
+            retrievedCount: resources.resources.length,
```
Garma00: Here we are tracking the single malformed message metadata, I think it is better to track informations abut the single record instead of `maximumID`, `minimumID`, `pageSize` and `retrievedCount`.

I would track the `messageId` only.

Garma00 (changes requested):

pnpm-workspace.yaml
robbcocco: Resolved in e05782d

giamma1295 (approved):

Garma00 (approved):

PR #859, opened 2026-07-16
Add tsx as dev dependency to avoid npx warning during local development

Garma00:


Garma00:
This PR removes annoying warnings during local development.
image

giamma1295 (approved):

PR #875, opened 2026-07-23
Remove documentdb dependency

Garma00:
This dependency is only used to import a type in a unit tests and it is causing the import of an old version of `underscore`.

image

By removing this dependency we  remove the old `underscore` version as well.

image

Closes IOCOM-3282


apps/services-func/src/functions/GetMessage/__tests__/handler.test.ts
```diff
@@ -61,6 +60,11 @@ import {
 } from "../../../generated/pagopa-ecommerce/PaymentDuplicatedStatusFaultPaymentProblemJson";
 import { GetMessageHandler } from "../handler";
 
+interface QueryError {
+  body: string;
+  code: number;
```
giamma1295: Apparently `QueryError` was mistakenly used here to mock the [MessageStatusModel](https://github.com/pagopa/io-functions-commons/blob/master/src/models/message_status.ts), which is returning CosmosError instead for both [findLastVersionByModelId](https://github.com/pagopa/io-functions-commons/blob/85d7f2425bad88f50fe9da6c8a3c5cae47fa4058/src/utils/cosmosdb_model_versioned.ts#L188) and [upsert](https://github.com/pagopa/io-functions-commons/blob/85d7f2425bad88f50fe9da6c8a3c5cae47fa4058/src/utils/cosmosdb_model_versioned.ts#L141)!

So we can remove it, ad update test accordingly!

giamma1295 (changes requested):

apps/services-func/src/functions/GetMessage/__tests__/handler.test.ts
Garma00: [11861f0](https://github.com/pagopa/io-messages/pull/875/commits/11861f0e30a2d93b19403c2f44cf10c856be9879)

giamma1295 (approved):
LGTM

PR #881, opened 2026-07-24
Add rc call in get messages

Garma00:
`getMessagesByUser`  now resolves the remote content configuration for messages carrying  `third_party_data`  by calling the  `rc-app`  micro-service, and computes the  `has_precondition`  flag from it as a fallback when the message itself does not specify it.

The implementation provides per-request cache to avoid duplicate  `rc-app`  calls for the same configuration id.

Closes IOCOM-3290


apps/messages-app/src/adapters/outbound/rc-confguration/rc-configuration.ts
```diff
@@ -0,0 +1,144 @@
...
+
+      if (!parsedResult.success)
+        return err(
+          new GenericError(
+            `malformed remote content configuration returned by the rc-app: ${parsedResult.error.message}`,
+          ),
+        );
+
+      return ok(this.toDomainRCConfiguration(parsedResult.data));
+    }
```
daniacca, line 115: Why don't put this also inside the `switch` with a `case 200:`?

infra/resources/_modules/container_apps/messages.tf
```diff
@@ -28,9 +28,18 @@ module "messages_ca" {
...
-        COMMON_COSMOS_URI               = var.common_cosmos_account.endpoint
-        COMMON_STORAGE_ACCOUNT_URI      = var.common_storage_account.endpoint
-        PN_SERVICE_ID                   = "01G40DWQGKY5GRWSNM4303VNRP" # PN
+        RC_APP_BASE_URL                 = "https://${module.remote_content_ca.url}/api/internal/rc-configurations"
+        SERVICE_TO_RC_MAP = jsonencode({
+          "01G40DWQGKY5GRWSNM4303VNRP" = "01HMVMHCZZ8D0VTFWMRHBM5D6F", # PN
+          "01GQQZ9HF5GAPRVKJM1VDAVFHM" = "01HMVMDTHXCESMZ72NA701EKGQ", # IO Sign
+          "01H4ZJ62C1CPGJ0PX8Q1BP7FAB" = "01HMVMCDD3JFYTPKT4ZN4WQ73B", # PagoPA Receipt (Test)
+          "01HD63674XJ1R6XCNHH24PCRR2" = "01HMVM9W74RWH93NT1EYNKKNNR", # PagoPA Receipt
+          "01GQQDPM127KFGG6T3660D5TXD" = "01HMVM4N4XFJ8VBR1FXYFZ9QFB", # Third Party Mock
```
daniacca, line 37: Can we move this map content so that there is no hardcoded values there?

daniacca (approved):
Good work! You can take a look at my comments, but they are not mandatory

infra/resources/_modules/container_apps/messages.tf
Garma00: Where do you propose to move this map? This already is an environmentvariable

apps/messages-app/src/adapters/outbound/rc-confguration/rc-configuration.ts
Garma00: Agree, I did this here [f3fdd44](https://github.com/pagopa/io-messages/pull/881/commits/f3fdd447476c69ac535b979f299d24cf0aa0aa2b)

Garma00: @daniacca unfortunately I had to [revert](https://github.com/pagopa/io-messages/pull/881/commits/64598a00c394f3235bd2e1dbb428a751a533a76e) this change due to a linter error https://github.com/pagopa/io-messages/actions/runs/30097573439/job/89495298310?pr=881#step:8:32.

daniacca: @Garma00 this happens because `case` don't create a scoped context by default. You could simply create a scoped context inside the case like this:

```typescript
case 200: {
      const jsonResponse = await ResultAsync.fromPromise(
        response.value.json(),
        () => new GenericError("invalid json response from rc-app"),
      );

      if (jsonResponse.isErr()) {
        return err(jsonResponse.error);
      }

      const parsedResult = rcConfigurationApiResponseSchema.safeParse(
        jsonResponse.value,
      );

      if (!parsedResult.success)
        return err(
          new GenericError(
            `malformed remote content configuration returned by the rc-app: ${parsedResult.error.message}`,
          ),
        );

      return ok(this.toDomainRCConfiguration(parsedResult.data));
}
break;
```

PR #893, opened 2026-07-29
Update Vitest to v4

Garma00:
Bumps  `vitest`  and  `@vitest/coverage-v8`  from  `^3.2.4`  to  `^4.1.0` in the  `pnpm-workspace.yaml`  catalog, and fixes the tests broken by the major upgrade.

All six come from the [Vitest 4 migration guide](https://vitest.dev/guide/migration.html#migration-guide):

- **dist/  is no longer excluded by default**: Compiled tests under  `dist/`  were being collected. Restored the exclude in the two  `packages/`  configs and added the missing configs for  `messages-app`  and  `rc-app`. (Simplified  exclude)
- **vi.spyOn  now returns the already-installed mock**: Fetch  call counts leaked across tests. Added  `beforeEach(() => vi.restoreAllMocks())`. (Changes to Mocking)
- **Mocks called with  new  now really construct**: Arrow implementations threw  is not a constructor , switched to  function. (spyOn and fn Support Constructors)
- **mockReturnValue  on a mock used with  new  is a hard error**: Replaced with  mockImplementation . (Mock class support)
- **test(name, fn, options)  signature removed**: Moved options to the second argument. (Deprecated APIs are Removed)
- **Stricter types**:  `ReturnType<typeof vi.fn>` and  as never  casts no longer typecheck, replaced with targeted casts.

Closes IOCOM-3285

### DIFF
--- a/apps/citizen-func/src/functions/GetMessages/__tests__/handler.test.ts
+++ b/apps/citizen-func/src/functions/GetMessages/__tests__/handler.test.ts
@@ -162,16 +162,15 @@ const aRetrievedMessageView: RetrievedMessageView = pipe(
 //----------------------------
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-const getByMessageContentByIdMock: ReturnType<typeof vi.fn> = vi
-  .fn()
-  .mockResolvedValue({
-    markdown: "a markdown",
-    subject: "a subject",
-  });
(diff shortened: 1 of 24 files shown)


daniacca (approved):
LGFM

giamma1295 (approved):

PR #903, opened 2026-08-03
Implement services-cms http adapter in get messages

robbcocco:
Closes IOCOM-3295

### DIFF
--- a/apps/messages-app/src/application/use-cases/get-user-messages.use-case.ts
+++ b/apps/messages-app/src/application/use-cases/get-user-messages.use-case.ts
@@ -30,6 +30,7 @@ import {
   RCConfiguration,
   RCConfigurationRepository,
 } from "../ports/rc-configuration.js";
+import { ServicesCmsRepository } from "../ports/services-cms.js";
 
 export type GetMessagesByUserUseCase = UseCase<
   {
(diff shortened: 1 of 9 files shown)


apps/messages-app/src/application/use-cases/get-user-messages.use-case.ts
```diff
@@ -356,7 +358,21 @@ export const makeGetMessagesByUserUseCase =
       return err(messageContents.error);
     }
 
+    const serviceIDs = [
+      ...new Set(selectedMessages.map((s) => s.metadata.senderServiceId)),
+    ];
+
+    const messageDetails =
```
Garma00:
```suggestion
    const serviceDetails =
```

apps/messages-app/src/application/use-cases/get-user-messages.use-case.ts
```diff
@@ -313,6 +314,7 @@ export const makeGetMessagesByUserUseCase =
     messageMetadataRepository: MessageMetadataRepository,
     messageStatusRepository: MessageStatusRepository,
     messageContentRepository: MessageContentRepository,
+    messageDetailRepository: ServicesCmsRepository,
```
Garma00:
```suggestion
    serviceDetailRepository: ServicesCmsRepository,
```

Garma00 (changes requested):

daniacca (approved):

Garma00 (approved):

apps/messages-app/src/application/use-cases/get-user-messages.use-case.ts
robbcocco: Resolved in c06a4c0

apps/messages-app/src/application/use-cases/get-user-messages.use-case.ts
robbcocco: Resolved in c06a4c0

PR #907, opened 2026-08-06
Remove rc code duplication

Garma00:
This PR removes the Remote Content's type duplication by moving it into `io-messages-common`.

It also fixes the returned values from the GetRCConfiguration by converting the properties from camelCase to snake_case as it is declared in the OpenAPI spec.


giamma1295 (approved):
LGTM

PR #918, opened 2026-08-26
Add getMessage to messages-app

lucamancosu:
Closes IOCOM-3319

### DIFF
--- a/apps/messages-app/src/application/ports/messages.ts
+++ b/apps/messages-app/src/application/ports/messages.ts
@@ -9,6 +9,7 @@
 // own types, so the application core does not depend on the HTTP/Fastify layer.
 import z from "zod";
 
+import { messageContentSchema } from "./message-content.js";
 import { messageStatusValueSchema } from "./message-status.js";
 
 const messageStatusAttributesSchema = z.object({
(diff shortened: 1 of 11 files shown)


apps/messages-app/src/application/ports/messages.ts
```diff
@@ -34,42 +35,64 @@ const thirdPartyDataSchema = z.object({
   original_receipt_date: z.string().optional(),
   original_sender: z.string().min(1).optional(),
   summary: z.string().min(1).optional(),
-});
-
-const messageCategoryPNSchema = thirdPartyDataSchema.extend({
```
Garma00, line 39: Why are we removing it?

apps/messages-app/src/application/use-cases/get-message.use-case.ts
```diff
@@ -0,0 +1,203 @@
...
+      new GenericError(`Cannot find service details for service ${serviceId}`),
+    );
+  }
+
+  return service.isErr()
+    ? err(new GenericError(service.error.message))
+    : service;
+};
+
+const computeCategory = (
```
Garma00, line 64: Maybe we can reuse the `computeMessageCategory` that we have in get-user-messages file

apps/messages-app/src/application/use-cases/get-message.use-case.ts
```diff
@@ -0,0 +1,203 @@
...
+    if (!publicMessage) return ok({ message });
+
+    const statusResult =
+      await messageStatusRepository.getLatestMessageStatusById(messageId);
+    if (statusResult.isErr()) {
+      return err(new GenericError(statusResult.error.message));
+    }
+
+    if (!service) {
+      const serviceResult = await getService(
```
Garma00, line 178: Can we avoid calling 2 times the getService method in this use-case?

Garma00 (changes requested):

apps/messages-app/src/application/use-cases/get-message.use-case.ts
lucamancosu: it was my first idea but i have preferred to duplicate it just for keep use cases isolated and if in the futurewe need to do a change for the computeCategory only on one use case we dont need to create a switch. 

If you prefer to create an utility or export the computeMessageCategory from get-user-messages i think there is  no problem. 

What you think ?

apps/messages-app/src/application/use-cases/get-message.use-case.ts
lucamancosu: Screenshot 2026-08-28 at 14 34 42

I compared this with the behavior in citizen-func. The current messages-app code handles the use case correctly. Calling getService unconditionally would have a downside: in some cases, the service data is not needed, and the additional call could introduce an unnecessary point of failure.

If you agree, I’ll leave the use case as it is.

apps/messages-app/src/application/ports/messages.ts
lucamancosu: i've did only beacuse it was a single zod line declaration and i have included it in the third party schema. If is possible that in the future we need to export the messageCategoryPNSchema i can revert it.

apps/messages-app/src/application/use-cases/get-message.use-case.ts
Garma00: I would prefer to extract this method and move it maybe on io-messages-common but I do not have a strong opinion, you can leave it as-is too.

Garma00 (approved):

PR #922, opened 2026-08-31
Add create message endpoint

Garma00:
Port the CreateMessage endpoint in the messages-app micro-service.

Closes IOCOM-3322


apps/messages-app/src/app.ts
```diff
@@ -119,6 +130,29 @@ export const createApp = (
     logger,
   );
 
+  const processingMessagePayloadStore = new BlobProcessingMessagePayloadStore(
+    commonStorageAccountClient,
+    config.PROCESSING_MESSAGE_CONTAINER_NAME,
+  );
```
giamma1295, line 146: This client is related to `iopstapi`, is this the right storage account for the `processing-message` blob?

Garma00: Nope, nice catch, resolved [here](https://github.com/pagopa/io-messages/pull/922#:~:text=Verified-,6ad7d13,-This%20branch%20has).

apps/messages-app/src/app.ts
```diff
@@ -141,6 +185,17 @@ export const createApp = (
       ),
     ]),
   );
+  mountCreateMessageHandler(
+    server,
+    makeCreateMessageUseCase(
+      messageMetadataCosmosAdapter,
+      servicesCmsAdapter,
+      remoteContentConfigurationReposiory,
```
giamma1295: Typo lack of `t` on word `Repository`, should be `remoteContentConfigurationRepository`.

Garma00: Nice catch, resolved [here](https://github.com/pagopa/io-messages/pull/922#:~:text=Verified-,d18bb1f,-This%20branch%20has).

apps/messages-app/src/application/use-cases/create-message.use-case.ts
```diff
@@ -0,0 +1,403 @@
...
+
+    return err(
+      new GenericError(
+        `invalid service detail for subscription ${subscriptionId}: ${serviceDetail.error.message}`,
+      ),
+    );
+  }
+
+  return ok(serviceDetail.value);
+};
```
giamma1295, line 159: The `servicesCmsRepository` provides two methods for retrieving services one for retrieving a single service and the other for retrieving a set of services. 
In this case, it seems that the single-service method (retrieving just one service) would be the most appropriate, but the method for retrieving multiple services is being used instead.

apps/messages-app/src/adapters/outbound/services-cms/services-cms.ts
giamma1295, line 178: This method is currently used only into the `create-message.use-case` to get a single service, is really needed or we can leave only the single one?

apps/messages-app/src/adapters/inbound/fastify/middlewares/user-email.middleware.ts
```diff
@@ -0,0 +1,42 @@
...
+  EmptyHttpMiddlewareContext,
+  HttpRequestMiddleware,
+} from "@pagopa/hexagonal-core";
+
+import { GenericError } from "@pagopa/hexagonal-core";
+import { err, ok } from "neverthrow";
+import z from "zod";
+
+const userEmailSchema = z.email().brand<"UserEmail">();
+export type UserEmail = z.TypeOf<typeof userEmailSchema>;
```
giamma1295: lib `hexagonal-core` contains this brandedType,  https://github.com/pagopa/io-platform-core/blob/main/packages/hexagonal-core/src/domain/value-objects/emailAddress.value-object.ts

Garma00: Nice catch, resolved [here](https://github.com/pagopa/io-messages/pull/922#:~:text=Verified-,5d44eb3,-Expose%20the%20single).

apps/messages-app/src/application/use-cases/create-message.use-case.ts
Garma00: Actually the `serviceCmsRepository` does not expose this method but it is only a private one in the adapter. [Here](https://github.com/pagopa/io-messages/pull/922/commits/f89340434c4ed45017c1ed5e4630364906d72109) I expose the method and use it in the create message as you suggested, thanks!

apps/messages-app/src/adapters/outbound/services-cms/services-cms.ts
Garma00: See [here](https://github.com/pagopa/io-messages/pull/922#discussion_r3923235253).

giamma1295 (approved):
LGTM

PR #933, opened 2026-09-09
Add list rcconfiguration endpoint

robbcocco:
Closes IOCOM-3304


Garma00 (changes requested):
We need to add the permissions to access the cosmos container.

Garma00 (approved):

